Issue #8, opened 2019-12-03
Add check: Incorrect delimiter


mgrachev:
Example:

```env
# .env

# Wrong
RAILS-ENV=development

# Correct
RAILS_ENV=development
```

This check should display a message:
> .env:1 The RAILS-ENV key has incorrect delimiter

sonro:
Can I have a crack at this?

mgrachev:
@sonro Sure! 👍

PR #20, opened 2019-12-15
Add check: Incorrect delimiter

sonro:
Addresses issue: #8 
I had to change the `checkout` function in src/check.rs - In order for different types to be bundled into a `Vec`, it needs to be wrapped in a `Box` which implements the `Check` trait.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,14 +27,25 @@ Detects if a line starts with a space or a tab character:
 DEBUG_HTTP=true
 ```
 
+### Incorrect delimiter
+
+Detects if a key does not use an underscore to sepearte words:
+```env
+# Wrong
+DB-NAME=testing
+
+# Correct
+DB_NAME=test
+```
+
 ## Roadmap
 - [ ] Add more checks:
   - [x] Leading Space
+  - [x] Incorrect delimiter
   - [ ] [Unordered keys](https://github.com/mgrachev/dotenv-linter/issues/4);
   - [ ] [Duplicated keys](https://github.com/mgrachev/dotenv-linter/issues/5);
   - [ ] [Lowercase keys](https://github.com/mgrachev/dotenv-linter/issues/6);
   - [ ] [Keys without values](https://github.com/mgrachev/dotenv-linter/issues/7);
-  - [x] Incorrect delimiter
   - [ ] [Spaces before or after the character `=`](https://github.com/mgrachev/dotenv-linter/issues/9);
   - [ ] Other checks.
 - [ ] Support [reviewdog](https://github.com/reviewdog/reviewdog);
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ DEBUG_HTTP=true
   - [ ] [Duplicated keys](https://github.com/mgrachev/dotenv-linter/issues/5);
   - [ ] [Lowercase keys](https://github.com/mgrachev/dotenv-linter/issues/6);
   - [ ] [Keys without values](https://github.com/mgrachev/dotenv-linter/issues/7);
-  - [ ] [Incorrect delimiter](https://github.com/mgrachev/dotenv-linter/issues/8);
+  - [x] Incorrect delimiter
   - [ ] [Spaces before or after the character `=`](https://github.com/mgrachev/dotenv-linter/issues/9);
   - [ ] Other checks.
 - [ ] Support [reviewdog](https://github.com/reviewdog/reviewdog);
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ RAILS_ENV=development
 
 ### Incorrect delimiter
 
-Detects if a key does not use an underscore to sepearte words:
+Detects if a key does not use an underscore to separate words:
 ```env
 # Wrong
 DB-NAME=testing
--- a/src/checks.rs
+++ b/src/checks.rs
@@ -30,7 +30,7 @@ trait Check {
 fn checklist() -> Vec<Box<dyn Check>> {
     vec![
         Box::new(leading_space::LeadingSpaceChecker::default()),
-        Box::new(incorrect_delimiter::IncorrectDelimiterChecker {}),
+        Box::new(incorrect_delimiter::IncorrectDelimiterChecker::default()),
     ]
 }
 
--- a/src/checks.rs
+++ b/src/checks.rs
@@ -1,6 +1,7 @@
 use crate::LineEntry;
 use std::fmt;
 
+mod incorrect_delimiter;
 mod leading_space;
 
 #[derive(Debug, Clone, PartialEq)]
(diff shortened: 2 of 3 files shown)


src/checks.rs
```diff
@@ -26,8 +27,11 @@ trait Check {
     fn run(&self, line: &LineEntry) -> Option<Warning>;
 }
 
-fn checklist() -> Vec<impl Check> {
-    vec![leading_space::LeadingSpaceChecker::default()]
+fn checklist() -> Vec<Box<dyn Check>> {
+    vec![
+        Box::new(leading_space::LeadingSpaceChecker::default()),
+        Box::new(incorrect_delimiter::IncorrectDelimiterChecker {}),
```
mgrachev: Please, implement the trait `Default` for this struct and use the method `default`:
```suggestion
        Box::new(incorrect_delimiter::IncorrectDelimiterChecker::default()),
```

src/checks/incorrect_delimiter.rs
```diff
@@ -0,0 +1,88 @@
+use crate::checks::{Check, Warning};
+use crate::LineEntry;
+
+pub(crate) struct IncorrectDelimiterChecker {}
```
mgrachev: This struct should have the field `warning: Warning`.
Example: https://github.com/mgrachev/dotenv-linter/blob/master/src/checks/leading_space.rs#L5

src/checks/incorrect_delimiter.rs
```diff
@@ -0,0 +1,88 @@
...
+
+impl Check for IncorrectDelimiterChecker {
+    fn run(&self, line: &LineEntry) -> Option<Warning> {
+        if line.raw_string.starts_with(' ') {
```
mgrachev: I think you shoud not do this check because it's the work of another linter - https://github.com/mgrachev/dotenv-linter/blob/master/src/checks/leading_space.rs.

src/checks/incorrect_delimiter.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn working_run() {
+        let checker = IncorrectDelimiterChecker {};
```
mgrachev:
```suggestion
        let checker = IncorrectDelimiterChecker::default();
```

src/checks/incorrect_delimiter.rs
```diff
@@ -0,0 +1,88 @@
...
+        let line = &LineEntry {
+            number: 1,
+            raw_string: String::from("DEBUG_HTTP=true"),
+        };
+        assert_eq!(None, checker.run(line));
+    }
+
+    #[test]
+    fn failing_run() {
+        let checker = IncorrectDelimiterChecker {};
```
mgrachev:
```suggestion
        let checker = IncorrectDelimiterChecker::default();
```

src/checks/incorrect_delimiter.rs
```diff
@@ -0,0 +1,88 @@
...
+            number: 1,
+            raw_string: String::from("DEBUG-HTTP=true"),
+        };
+        let expected = Some(Warning::new("The DEBUG-HTTP key has incorrect delimiter"));
+        assert_eq!(expected, checker.run(line));
+    }
+
+    #[test]
+    fn unformated_run() {
+        let checker = IncorrectDelimiterChecker {};
```
mgrachev:
```suggestion
        let checker = IncorrectDelimiterChecker::default();
```

src/checks/incorrect_delimiter.rs
```diff
@@ -0,0 +1,88 @@
...
+        let line = &LineEntry {
+            number: 1,
+            raw_string: String::from("DEBUG-HTTPtrue"),
+        };
+        assert_eq!(None, checker.run(line));
+    }
+
+    #[test]
+    fn leading_space_run() {
+        let checker = IncorrectDelimiterChecker {};
```
mgrachev:
```suggestion
        let checker = IncorrectDelimiterChecker::default();
```

src/checks/incorrect_delimiter.rs
```diff
@@ -0,0 +1,88 @@
...
+        let line = &LineEntry {
+            number: 1,
+            raw_string: String::from(" DEBUG_HTTP=true"),
+        };
+        assert_eq!(None, checker.run(line));
+    }
+
+    #[test]
+    fn empty_run() {
+        let checker = IncorrectDelimiterChecker {};
```
mgrachev:
```suggestion
        let checker = IncorrectDelimiterChecker::default();
```

src/checks/incorrect_delimiter.rs
```diff
@@ -0,0 +1,88 @@
...
+        let line = &LineEntry {
+            number: 1,
+            raw_string: String::from(""),
+        };
+        assert_eq!(None, checker.run(line));
+    }
+
+    #[test]
+    fn short_run() {
+        let checker = IncorrectDelimiterChecker {};
```
mgrachev:
```suggestion
        let checker = IncorrectDelimiterChecker::default();
```

src/checks/incorrect_delimiter.rs
sonro: If this check is not included this linter returns false posititives for any `key` that mistakenly uses `" "` as a delimiter.

README.md
```diff
@@ -34,7 +34,7 @@ DEBUG_HTTP=true
   - [ ] [Duplicated keys](https://github.com/mgrachev/dotenv-linter/issues/5);
   - [ ] [Lowercase keys](https://github.com/mgrachev/dotenv-linter/issues/6);
   - [ ] [Keys without values](https://github.com/mgrachev/dotenv-linter/issues/7);
-  - [ ] [Incorrect delimiter](https://github.com/mgrachev/dotenv-linter/issues/8);
+  - [x] Incorrect delimiter
```
mgrachev: Please, move it up and add an example like this https://github.com/mgrachev/dotenv-linter/blob/master/README.md#leading-space

src/checks/incorrect_delimiter.rs
mgrachev: I see. Maybe for that it's better to use method `is_whitespace()`?
```rust
if key.chars().any(|c| !c.is_alphabetic() && !c.is_whitespace() && c != '_') {}
```

sonro: Ah good suggestion. Although, that would mean the linter would report `SOME KEY=true` as being correct. If this is acceptable behaviour I will action this change.

mgrachev: Good catch! It's not acceptable behaviour. Can you add a test for that?

mgrachev: But now this check returns some warnings if I have lines like those:
```env
ABCD= 1
ABCD =2
ABCD = 3
```
It's the work another linter - https://github.com/mgrachev/dotenv-linter/issues/9.
Could you fix it and cover with tests?

README.md
```diff
@@ -41,14 +41,25 @@ RAILS_ENV=
 RAILS_ENV=development
 ```
 
+### Incorrect delimiter
+
+Detects if a key does not use an underscore to sepearte words:
```
```suggestion
Detects if a key does not use an underscore to separate words:
```

mgrachev:
@sonro Thank you very much! You're awesome! ❤️

PR #21, opened 2019-12-16
Lowercase checker

qelphybox:
Closes https://github.com/mgrachev/dotenv-linter/issues/6

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,6 @@ $ ./dotenv-linter
 ### Leading Space
 
 Detects if a line starts with a space or a tab character:
-
 ```env
 # Wrong
  DEBUG_HTTP=true
@@ -31,7 +30,6 @@ DEBUG_HTTP=true
 ### Keys Without Values
 
 Detects if a line has a key without a value:
-
 ```env
 # Wrong
 RAILS_ENV
@@ -46,7 +44,6 @@ RAILS_ENV=development
 ### Incorrect delimiter
 
 Detects if a key does not use an underscore to separate words:
-
 ```env
 # Wrong
 DB-NAME=testing
@@ -59,20 +56,21 @@ DB_NAME=test
 
 Detects if key has lowercase characters
 
-```` `env
+```env
 # Wrong
 DEbUG_hTTP=true
 debug_http=true
 
 # Correct
 DEBUG_HTTP=true
+```
 
 ## Roadmap
 - [ ] Add more checks:
   - [x] Leading Space;
   - [x] Keys without values;
   - [x] Incorrect delimiter;
-  - [x] Lowercase keys;
+  - [x] Lowercase keys
   - [ ] [Unordered keys](https://github.com/mgrachev/dotenv-linter/issues/4);
   - [ ] [Duplicated keys](https://github.com/mgrachev/dotenv-linter/issues/5);
   - [ ] [Spaces before or after the character `=`](https://github.com/mgrachev/dotenv-linter/issues/9);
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -27,12 +27,24 @@ Detects if a line starts with a space or a tab character:
 DEBUG_HTTP=true
 ```
 
+### Lowercase key
+
+Detects if key has lowcase characters
```
```suggestion
Detects if key has lowercase characters
```

src/checks.rs
```diff
@@ -26,8 +27,11 @@ trait Check {
     fn run(&self, line: &LineEntry) -> Option<Warning>;
 }
 
-fn checklist() -> Vec<impl Check> {
-    vec![leading_space::LeadingSpaceChecker::default()]
+fn checklist() -> Vec<Box<dyn Check>> {
+    vec![
+        Box::new(leading_space::LeadingSpaceChecker::default()),
+        Box::new(lowercase_key::LowercaseKeyChecker {}),
```
mgrachev: Please, implement the `trait Default` for this struct and use the method `default`:
```suggestion
        Box::new(lowercase_key::LowercaseKeyChecker::default()),
```

src/checks/lowercase_key.rs
```diff
@@ -0,0 +1,51 @@
+use crate::checks::{Check, Warning};
+use crate::LineEntry;
+
+pub(crate) struct LowercaseKeyChecker {}
```
mgrachev: This struct should have the field `warning: Warning`.
Example: https://github.com/mgrachev/dotenv-linter/blob/master/src/checks/leading_space.rs#L5

src/checks/lowercase_key.rs
```diff
@@ -0,0 +1,51 @@
...
+
+impl Check for LowercaseKeyChecker {
+    fn run(&self, line: &LineEntry) -> Option<Warning> {
+        let line_str: Vec<&str> = line.raw_string.split('=').collect();
+        let key = line_str[0];
+        if key.to_uppercase() == key {
```
mgrachev: The method `to_uppercase()` returns `String`, but the variable `key` is `&str`.
I think you should convert `&str` to `String`.
```suggestion
        if key.to_uppercase() == key.to_string() {
```

src/checks/lowercase_key.rs
```diff
@@ -0,0 +1,51 @@
...
+            None
+        } else {
+            Some(Warning {
+                message: format!("The {} key should be in uppercase", key),
```
mgrachev: Please, use the field `warning` to store and retrieve a message:
```suggestion
                message: format!(self.warning.clone(), key),
```

src/checks/lowercase_key.rs
```diff
@@ -0,0 +1,51 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn lowercase_key_checker_run() {
+        let checker = LowercaseKeyChecker {};
```
mgrachev: Please, use the `trait Default`:
```suggestion
        let checker = LowercaseKeyChecker::default();
```

mgrachev:
@qelphybox  👋

Thank you for your contribution! ❤️

I've left a few comments. Could you please fix them?

qelphybox:
thx for review, i'll fix it soon

src/checks/lowercase_key.rs
qelphybox: sorry, It seems kinda weird to me create warning with message template: 
```rust
pub(crate) struct LowercaseKeyChecker {
    warning: Warning
}

// like this
impl Default for LowercaseKeyChecker {
    fn default() -> Self {
        Self {
            warning: Warning::new("The {} key should be in uppercase"),
        }
    }
}

// then i should extract message template from warning
//...
Some(Warning {
    message: format!(self.warning.clone(), key) // btw it doesnt work: format! receives string literal on the first place
})
```
can you explain, why we should create struct with warning, only for storing message template, why we can't store only message? :thinking:

src/checks/lowercase_key.rs
qelphybox: thx, i think it will be better
```rust
let key = line_str[0].to_string();
if key.to_uppercase() == key {
```

src/checks/lowercase_key.rs
qelphybox: https://github.com/mgrachev/dotenv-linter/pull/21#discussion_r358616840

src/checks/lowercase_key.rs
mgrachev: Yes, you're right. I missed that. What do you think about this solution?
```rust
pub(crate) struct LowercaseKeyChecker {
    template: String
}

impl Default for LowercaseKeyChecker {
    fn default() -> Self {
        Self {
            warning: String::from("The {} key should be in uppercase"),
        }
    }
}

Some(Warning {
    message: Some(Warning { message: self.template.replace("{}", key) })
})
```

qelphybox: seems ok, i'll take it

qelphybox:
done, rereview pls

mgrachev:
@qelphybox Thank you very much! You're awesome! ❤️

PR #22, opened 2019-12-16
Add key without value check

### LINKED ISSUE #7: Add check: Keys without values
Example:

```env
# .env

# Wrong
RAILS_ENV

# Correct
RAILS_ENV=

# Correct
RAILS_ENV=development
```

This check should display a message:
> .env:1 The RAILS_ENV key should be with a value

mstruebing:
closes #7

### DIFF
--- a/src/checks/key_without_value.rs
+++ b/src/checks/key_without_value.rs
@@ -1,0 +1,51 @@
+use crate::checks::{Check, Warning};
+use crate::LineEntry;
+
+pub(crate) struct KeyWithoutValueChecker {
+    warning: Warning,
+}
+
+impl Default for KeyWithoutValueChecker {
+    fn default() -> Self {
+        Self {
+            warning: Warning::new("Key without value detected"),
+        }
+    }
+}
+
+impl Check for KeyWithoutValueChecker {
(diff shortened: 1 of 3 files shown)


mstruebing:
Thank you, it was a pleasure :)
I've fixed your suggestions and added the check also to the readme, I also removed the link to the issue (I've thought this would be want you want)

src/checks/key_without_value.rs
```diff
@@ -0,0 +1,51 @@
...
+        Self {
+            warning: Warning::new("The {} key should be with a value or have an equal sign"),
+        }
+    }
+}
+
+impl Check for KeyWithoutValueChecker {
+    fn run(&self, line: &LineEntry) -> Option<Warning> {
+        if !line.raw_string.contains('=') {
+            Some(self.warning.clone())
```
mgrachev: You should find a key and substitute it in a message to get the message like that: `The RAILES_ENV key should be with a value or have an equal sign`

mstruebing: Fixed. :)

mstruebing:
:tada:

mgrachev:
@mstruebing Thank you very much! You're awesome! ❤️

Issue #47, opened 2020-01-05
Exit with the code 1 If there is at least one warning


mgrachev:
Link: https://github.com/mgrachev/dotenv-linter/blob/master/src/lib.rs#L49

mstruebing:
Hey, I think it would be better if the tool is failing with an exit code when at least one warning exists by default and not with an additional argument.
This is default behavior of a linter in my opinion.

mgrachev:
@mstruebing Good catch! Thank you. I've updated this issue.

mstruebing:
One question, why should the user be capable of manually choosing which exit code is used when failing occurs?
As far as I understand exit codes, and used them, it is totally up to the tool to decide which exit code to use.
Every exit code has a meaning, so it shouldn't be user decidable.

I.e you could use an exit code 1 for example to indicate wrongly used arguments and exit code 2 for errors in the `.env`-files.

So in short: Exit code has meaning, shouldn't be customizable by user input.

mgrachev:
@mstruebing I think you're right.

But I have another question - Should we give the user the ability to change the exit code from 1 to 0 if dotenv-linter finds at least one warning?

For example, [golangci-lint](https://github.com/golangci/golangci-lint) lets do it like that:
```bash
$ golangci-lint run --issues-exit-code=0
```

Or another example - [golint](https://github.com/golang/lint) sets the exit status to 1 if any issues are found like that:
```bash
$ golint -set_exit_status
```

mstruebing:
I don't think this is a good idea. I run the linter to catch issues and get notified, this is done best by an `exit code != 0`.

What would make sense, in my opinion, is, if we would make a difference between warnings and errors (which we don't do at the moment, and I don't know if it would be cool) and give the user the option to decide if warnings would produce an `exit code != 0` and errors always produce an `exit code != 0`.

mgrachev:
Updated.

Louis-Lesage:
I have made a pull request that should solve this issue.
See #58 .
Feedback is appreciated

PR #58, opened 2020-01-09
Exit code on warnings

Louis-Lesage:
I added a verification on the amount of warnings found and exit with 1 as error code when the number or warnings is more than 0.

Should solve #47 .


mgrachev:
@Louis-Lesage 👋Thank you for your contribution!  🚀

Can you update the changelog file too?

mgrachev:
@Louis-Lesage Thanks! ❤️

PR #62, opened 2020-01-11
Fix multiple paths by dedup

### LINKED ISSUE #57: Fix a bug with duplicate paths
We can pass an argument to `dotenv-linter` to include some files to check:

```bash
$ dotenv-linter -i .my-env-file
.env:1 Leading space detected
.my-env-file:1 Leading space detected
```

If the file already exists in the paths, we will get a duplicate result:

```bash
$ dotenv-linter -i .env -i .env
.env:1 Leading space detected
.env:1 Leading space detected
.env:1 Leading space detected
```

Need to fix that and write tests for that case.

Link: https://github.com/mgrachev/dotenv-linter/blob/master/src/lib.rs#L101

mstruebing:
closes #57

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,26 +82,3 @@ jobs:
     - uses: actions-rs/cargo@v1
       with:
         command: build
-
-  integration:
-    name: Integration
-    runs-on: ubuntu-latest
-    steps:
-    - uses: actions/checkout@v1
-    - uses: actions-rs/toolchain@v1
-      with:
-        toolchain: stable
-    - name: Rust version
-      run: rustc --version
-    - uses: actions-rs/cargo@v1
-      with:
-        command: build
-    - name: Run Integration test
-      run: |
-        cd test 
-        if [ $(cargo run | wc -l) -ne 4 ] 
-        then 
-            exit 1 
-        else 
-          exit 0 
-        fi
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,3 +82,26 @@ jobs:
     - uses: actions-rs/cargo@v1
       with:
         command: build
+
+  integration:
+    name: Integration
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions-rs/toolchain@v1
+      with:
+        toolchain: stable
+    - name: Rust version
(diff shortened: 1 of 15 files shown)


src/lib.rs
```diff
@@ -22,7 +22,8 @@ pub struct LineEntry {
 }
 
 pub fn run(matches: clap::ArgMatches) -> Result<(), Error> {
-    let paths = dotenv_files(matches)?;
+    let mut paths = dotenv_files(matches)?;
+    paths.dedup();
```
mgrachev: It doesn't work for the case:
```bash
$ dotenv-linter -i .my-env-file -i .env
.env:1 Leading space detected
.my-env-file:1 The line has spaces around equal sign
.env:1 Leading space detected
```

mgrachev: I think you should use [HashSet](https://doc.rust-lang.org/std/collections/struct.HashSet.html) to fix that.

mgrachev:

I've left a few comments. Could you please fix them and write integration tests to check that this functionality works properly?

src/lib.rs
mstruebing: Ah yes, because `dedup` only removes duplicates which follow after each other. Will fix :)

mstruebing: A sort before the `dedup` would also work. But I think the HashSet is better.

.github/workflows/ci.yml
```diff
@@ -82,3 +82,26 @@ jobs:
     - uses: actions-rs/cargo@v1
       with:
         command: build
+
+  integration:
```
mgrachev: I mean you should write integration tests like that - https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html

For example: Create the directory `tests` and test the function `run` from `lib.rs`.

mgrachev: @mstruebing I've thought, probably I should write integration tests instead of you because I have some ideas. Don't you mind?

mstruebing: Yes sure, sorry for not pushing this forward :)

mgrachev:
@mstruebing Thank you! ❤️

src/lib.rs
mgrachev: @mstruebing `sort` and `dedup` is better 😃
https://github.com/mgrachev/dotenv-linter/pull/68

PR #63, opened 2020-01-11
Make leading space detector more robust and aware of allowed chars 

### LINKED ISSUE #56: Leading space check: Doesn't check tab character
Need to fix this bug and write a test for this case.

Link: https://github.com/mgrachev/dotenv-linter/blob/master/src/checks/leading_space.rs#L18

mstruebing:
I created this proof of concept to use a regular expression in the leading space checker.
However, as I'm not the maintainer I don't know if this is the way where this project wants to go.

If you would go for it we have two options I can currently see:

* rename the leading space checker to something it really reflects (leading character checker or something)
* Combine some other checkers which only act on one line also into a regex and one checker (for example, I think it would be very easy to combine the leading space, spaces around equal, and uppercase key checker, quick and dirty and not fully functional: `^[#A-Z_]*=.*`

closes #56

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ FOO=BAR
 
 ### Leading character
 
-Detects if a line starts with an allowed character (characters from a to z and `_` (underscore) are allowed)
+Detects if a line starts with an unallowed character (characters from `A` to `Z` and `_` (underscore) are allowed):
 
 ```env
 ❌Wrong
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ By default, `dotenv-linter` checks all files with the `.env` prefix. For example
 
 ```bash
 $ dotenv-linter
-.env:1 Leading space detected
+.env:1 Invalid leading character detected
 .env:2 The FOO-BAR key has incorrect delimiter
 .env:3 The FOo_BAR key should be in uppercase
 .env:4 The line has spaces around equal sign
@@ -68,7 +68,7 @@ you can use the argument `--include FILE_NAME` or its short version `-i FILE_NAM
 
 ```bash
 $ dotenv-linter -i test.env --include .my-env-file
-.env:1 Leading space detected
+.env:1 Invalid leading character detected
 test.env:2 The FOO-BAR key has incorrect delimiter
 .my-env-file:3 The line has spaces around equal sign
 ```
@@ -128,9 +128,9 @@ FOO=
 FOO=BAR
 ```
 
-### Leading space
+### Leading character
 
-Detects if a line starts with a space or a tab character:
+Detects if a line starts with an allowed character (characters from a to z and `_` (underscore) are allowed)
 
 ```env
 ❌Wrong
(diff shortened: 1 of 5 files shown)


mgrachev:

I need a bit more time to think about it.

src/checks/leading_space.rs
```diff
@@ -15,8 +17,10 @@ impl Default for LeadingSpaceChecker {
 
 impl Check for LeadingSpaceChecker {
     fn run(&mut self, line: LineEntry) -> Option<Warning> {
-        // FIXME: Doesn't check a tab character
-        if line.raw_string.starts_with(' ') {
+        // Allow Uppercase, Lowercase and # as first character
+        let re = Regex::new(r"^[A-Za-z#]").unwrap();
+
+        if !re.is_match(&line.raw_string) {
```
mgrachev: I think we can check it easier, without using the `regex` crate:
```rust
if !line.raw_string.starts_with(|c: char| c.is_alphabetic() || c == '_') {
    // return Warning
} else {
    // return None
}
```

@mstruebing What do you think about the example above?

mgrachev:
@mstruebing I thought about it and came to the following opinion.

In general, I like the idea to check valid characters instead of invalid characters, but I wouldn't like to use the `regex` crate to do that. Rust provides the standard library which allows doing the same without additional dependencies. You can see it in the example above.

> rename the leading space checker to something it really reflects (leading character checker or something)

mgrachev:
Ideally you need to make sure (write tests) that this check covers all of these cases:
```
// FOO=BAR => None
// _FOO=BAR => None
// .FOO=BAR => Warning
// *FOO=BAR => Warning
// 1FOO=BAR => Warning
// <tab>FOO=BAR => Warning
// <space>FOO=BAR => Warning
```

mstruebing:
I aagree with you, I will work on that at latest next week

mgrachev:
@mstruebing Thank you! Could you make some thanges?
* Update the warning message to "Invalid leading character detected" or something else (https://github.com/mgrachev/dotenv-linter/pull/63/files#diff-9b42d93a18412a9385bca014eb36fed5R11);
* Update the message commit because we don't use the `regex` crate to check wrong leading characters;
* Add this changes to `CHANGELOG.md`.

mstruebing:
Updated the PR title and docs, I also renmaed `leading_space` to `leading_character` to reflect more what it really is.
Hope this is fine

README.md
```diff
@@ -128,9 +128,9 @@ FOO=
 FOO=BAR
 ```
 
-### Leading space
+### Leading character
 
-Detects if a line starts with a space or a tab character:
+Detects if a line starts with an allowed character (characters from a to z and `_` (underscore) are allowed)
```
```suggestion
Detects if a line starts with an unallowed character (characters from `A` to `Z` and `_` (underscore) are allowed):
```

mgrachev:
@mstruebing Please update the [Roadmap](https://github.com/mgrachev/dotenv-linter#-roadmap) section in the README.md as well.

mgrachev:
@mstruebing Thanks a lot! You're cool! ❤️

PR #65, opened 2020-01-14
Add support for specified directory

Louis-Lesage:
Added cli argument --path or -p where you specify the path of the directory where you want dotenv-linter to run.

If there is no argument -p, the program use the current directory.


mgrachev:
@Louis-Lesage 👋 Thanks a lot! You're awesome! ❤️

PR #74, opened 2020-01-24
WIP Strip inline comments

### LINKED ISSUE #38: Remove inline comments from lines
We should remove inline comments before running checks, because inline comments may affect checks behaviour.

Example:
```env
FOO=BAR # some comment ===
```

Is need to ensure that commented lines with leading spaces will be skipped too:

```env
 # FOO=BAR
```

mstruebing:
closes #38


mgrachev:

I'm still not completely sure whether to do this. Can we discuss it there https://github.com/mgrachev/dotenv-linter/issues/38?

PR #88, opened 2020-02-24
Add `RELEASING.md`

### LINKED ISSUE #82: Provide dotenv-linter in the Arch User Repository
The Arch User Repository is like a package repository for Arch linux maintained by it's users.
I would be happy to set it up there so that any arch user can simply install it with:

`<your-aur-helper> -S dotenv-linter` (`-S` is mostly the command to install a package, in case your aur helper does it differently you need to substitute this command.

If that would be desired I would be happy to go :)

mstruebing:
Close #82


mgrachev:
@mstruebing 👋You've done a wonderful job! 🚀Thanks a lot! ❤️

PR #95, opened 2020-03-04
Fix digits incorrect delimiter

alter369:


src/checks/incorrect_delimiter.rs
```diff
@@ -16,7 +16,11 @@ impl Default for IncorrectDelimiterChecker {
 impl Check for IncorrectDelimiterChecker {
     fn run(&mut self, line: LineEntry) -> Option<Warning> {
         let key = line.get_key()?;
-        if key.trim().chars().any(|c| !c.is_alphabetic() && c != '_') {
+        if key
+            .trim()
+            .chars()
+            .any(|c| !c.is_alphabetic() && !c.is_numeric() && c != '_')
```
mgrachev: Rust has the method [is_alphanumeric](https://doc.rust-lang.org/std/primitive.char.html#method.is_alphanumeric) which uses these methods under the hood. Can we use it?

mgrachev:
@alter369 👋Thank you for your contribution! 🚀

I've left a comment. Can you fix it, as well as add this PR to [CHANGELOG.md](https://github.com/mgrachev/dotenv-linter/blob/master/CHANGELOG.md)?

src/checks/incorrect_delimiter.rs
alter369: Sure 👌🏻

mgrachev:
@alter369 Thanks a lot! 👍

PR #97, opened 2020-03-13
Fix --path CLI parameter

pineapplethief:
* Fixed --path CLI parameter handling not canonizing filepaths


mgrachev:
@pineapplethief 👋Thank you for your contribution! ❤️

PR #107, opened 2020-03-22
Exit with 0 when there are no warnings

### LINKED ISSUE #100: Fix a bug with exit with code 1
`dotenv-linter` always exits with code 1 even if there are no warnings.
Need to fix that bug and write some integration tests to cover that case.

Link: https://github.com/mgrachev/dotenv-linter/blob/master/src/main.rs#L18

boybird:
Resolves #100


mgrachev:
@boybird 👋Thank you for your contribution! 🚀

Sorry, but I have already merged the PR with these changes https://github.com/mgrachev/dotenv-linter/pull/105 😔

Issue #101, opened 2020-03-22
UnorderedKeys: Use Vector#get for getting value by index


mgrachev:
Need to replace the current solution of checking the length of the vector and getting a value by a key with the method [Vector#get](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.get). It's more clear and convenient way to do that.

Source: https://github.com/mgrachev/dotenv-linter/blob/master/src/checks/unordered_keys.rs#L28:L33

PR #108, opened 2020-03-22
UnorderedKeys: Use Vector#get for getting value by index

boybird:
Use `get` method to get result of item in `Vec` and use ? operator unwrap the result if it's `Some`.
Resolve: #101

### DIFF
--- a/src/checks/unordered_keys.rs
+++ b/src/checks/unordered_keys.rs
@@ -48,6 +48,7 @@ impl Check for UnorderedKeysChecker {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use std::path::PathBuf;
 
     fn run_unordered_tests(asserts: Vec<(LineEntry, Option<Warning>)>) {
         let mut checker = UnorderedKeysChecker::default();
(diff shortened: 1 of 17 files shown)


src/checks/unordered_keys.rs
```diff
@@ -25,12 +25,7 @@ impl Check for UnorderedKeysChecker {
         if !sorted_keys.eq(&self.keys) {
             let index = sorted_keys.iter().position(|p| p == &key)?;
 
-            if (index + 1) >= sorted_keys.len() {
-                // If the vectors are not equal, but the key is the last element
-                return None;
-            }
-
-            let another_key = sorted_keys[index + 1].clone();
+            let another_key = sorted_keys.get(index + 1)?.clone();
```
mgrachev:
```suggestion
            let another_key = sorted_keys.get(index + 1)?;
```
I think we shouldn't use the `clone` method, because the `get` method returns a reference.

mgrachev:
@boybird 👋Thank you for your contribution! 🚀

I have left a comment. Could you fix it and, as well as add this PR to [CHANGELOG.md](https://github.com/mgrachev/dotenv-linter/blob/master/CHANGELOG.md) (at the top of the list)?

mgrachev:
@boybird Thanks a lot! 👍

If it’s not difficult for you, please support the project - click on the star in this repository on GitHub ⭐️

PR #113, opened 2020-03-22
Use HashSet for DuplicateKeyChecker

TamasFlorin:
The current `DuplicateKeyChecker` implementation uses a `Vec<String>` to store previously seen keys. This leads to an `O(n)` time-complexity when we check if a key is already present in the keys collection.
In order to improve the time-complexity of this checker, we can use a `HashSet<String>` which has `O(1)` time-complexity (on average) for the `contains` operation.


mgrachev:
@TamasFlorin Thank you for your contribution! 💪

Could you update your branch from the master branch and then fix the merge conflict?

mgrachev:
@TamasFlorin Thanks a lot! You're awesome! ❤️

PR #138, opened 2020-04-02
Added DuplicateKey string to check name message.

### LINKED ISSUE #120: DuplicatedKey: Show check name in message
Right now the message looks like this:
```
.env:2 The FOO key is duplicated
```

Need to add the check name to the message:
```
.env:2 DuplicatedKey: The FOO key is duplicated
```

SaMuRa1ReM1X:
Fixes #120 

#### ✔ Checklist:

- [x] This PR has been added to [CHANGELOG.md](https://github.com/mgrachev/dotenv-linter/blob/master/CHANGELOG.md) (at the top of the list);


src/checks/duplicated_key.rs
```diff
@@ -11,7 +11,7 @@ impl Default for DuplicatedKeyChecker {
     fn default() -> Self {
         Self {
             keys: HashSet::new(),
-            template: String::from("The {} key is duplicated"),
+            template: String::from("DuplicatedKey: The {} key is duplicated"),
```
mgrachev, line 14: In the future I would like to have the ability to disable some checks by their names.
To do that, all checks should have the `name` field.
Please, move the check's name from the template to the special field.
Also, I would like to have a method to show the check's name in the message, for example:
```rust
impl DuplicatedKeyChecker {
    fn message(&self, key: &str) -> String {
      // code...
    }
}
```

mgrachev:
@SaMuRa1ReM1X Thanks a lot for your contribution! ❤️

I have left a comment. Could you please fix it?

src/checks/duplicated_key.rs
SaMuRa1ReM1X: I'll implement the message function as specified. Sorry for the delayed response.

mgrachev:
@SaMuRa1ReM1X Thank you! 🚀

Issue #123, opened 2020-03-27
LeadingCharacter: Show check name in message


mgrachev:
Right now the message looks like this:
```
.env:2 Invalid leading character detected
```

Need to add the check name to the message:
```
.env:2 LeadingCharacter: Invalid leading character detected
```

undef1nd:
Heya, I can give it a try :-)

michaetto:
How about this: https://github.com/mgrachev/dotenv-linter/pull/141 ?
Not sure why to make reallocations for static strings.
I think other checker can be changed this way as well.
Also maybe cloning will not be needed if Warning type etc. are also borrowed with proper lifetimes,  make reallocation when it is actually needed (most likely in replace scenarios etc.)

michaetto:
@undef1nd do not hesitate to proceed with your solution, I've just only wanted to raise the topic of unnecessary reallocation with #141 as an example

undef1nd:
@michaetto But yours actually solves the problem, right? I don't have any objections against it being merged :-)

mgrachev:
@michaetto Thank you for raising the topic of unnecessary reallocations. I totally agree with you.
If you know how to optimize mu code - let me know 😃 Any help appreciated! 🙏

As I can see - you have done this issue, so I will assign this issue to you if you don't mind.

mgrachev:
@undef1nd Sure!

michaetto:
All right then :+1: , you have it here: #144

PR #141, opened 2020-04-03
#123

michaetto:
#### ✔ Checklist:
- [ x] Tests for the changes have been added (for bug fixes / features);

### DIFF
--- a/src/checks/leading_character.rs
+++ b/src/checks/leading_character.rs
@@ -2,13 +2,13 @@ use crate::checks::Check;
 use crate::common::*;
 
 pub(crate) struct LeadingCharacterChecker {
-    template: String,
+    template: &'static str,
 }
 
 impl Default for LeadingCharacterChecker {
     fn default() -> Self {
         Self {
-            template: String::from("Invalid leading character detected"),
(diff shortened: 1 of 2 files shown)


src/checks/leading_character.rs
```diff
@@ -2,13 +2,13 @@ use crate::checks::Check;
...
 pub(crate) struct LeadingCharacterChecker {
-    template: String,
+    template: &'static str,
 }
 
 impl Default for LeadingCharacterChecker {
     fn default() -> Self {
         Self {
-            template: String::from("Invalid leading character detected"),
+            template: "LeadingCharacter: Invalid leading character detected",
```
mgrachev, line 11: In the future I would like to have the ability to disable some checks by their names.
To do that, all checks should have the name field.
Please, move the check's name from the template to the special field.
Also, I would like to have a method to show the check's name in the message, for example:
https://github.com/mgrachev/dotenv-linter/commit/38b9558bd38341ff13fa07f27fc069a9ce9aa48e#diff-b598dd141594bc9ca17caeca0ed82794R10

PR #169, opened 2020-05-01
No landing pads fix

evgeniy-r:
It seems that Rust team [removes](https://github.com/rust-lang/rust/pull/70175) a `no-landing-pads` compiler option yesterday.

And now there is an error in "Tests and coverage" job in CI workflow: "unknown debugging option: `no-landing-pads`".

I have fix it, but I am not sure of the correctness of this solution.


mgrachev:
@evgeniy-r Thank you 🦀

Issue #199, opened 2020-05-23
Add an argument to show available checks


mgrachev:
To skip some checks, we need to know which checks are available in the dotenv-linter.

```
$ dotenv-linter --show-checks
DuplicatedKey
EndingBlankLine
ExtraBlankLine
IncorrectDelimiter
KeyWithoutValue
LeadingCharacter
LowercaseKey
QuoteCharacter
SpaceCharacter
UnorderedKey
```

Issue #23, opened 2019-12-16
Ideas


mgrachev:
Hi there!

If you have any ideas on how to improve this project or what other checks we should add - welcome 😉

mstruebing:
A dockerimage would be nice I think

mgrachev:
> A dockerimage would be nice I think

Could you give more information about that? Do you want to run this linter using a Docker image?

mstruebing:
Yes, an image with an already compiled binary which I can just use without the need to install `dotenv-linter` or any dependencies (like rust for example).

qelphybox:
disabling comments
e.g.
```env
# linter:disable UnorderedKeysChecker
 
MY=val
OWN=val
ORDER=val
```

qelphybox:
We have an `.env`
```env
 LOGGER_LEVEL=info
RAILS_ENV
DB-NAME=testing
DEbUG_hTTP=true
DB_NAME=development

```

run `dotenv-linter --fix`

- remove leading spaces
- replace incorrect delimeter to "_"
- set '=' to empty keys
- remove duplicated keys, sets value of the last key   
- order keys

We'll get:
```env
DB_NAME=development
DEBUG_HTTP=true
LOGGER_LEVEL=info
RAILS_ENV=
```

mgrachev:
> Yes, an image with an already compiled binary which I can just use without the need to install `dotenv-linter` or any dependencies (like rust for example).

Great idea! I've just created an issue to do that https://github.com/mgrachev/dotenv-linter/issues/34.

mgrachev:
> Disabling/enabling comments. It's a good option, i think some users may want to disable some checkers
> e.g.
> 
> ```
> # linter:disable UnorderedKeysChecker
>  
> MY=val
> OWN=val
> ORDER=val
> ```

I think it's a good idea! Could you open an issue and write your vision on how this will work?

mgrachev:
> I think autofix option is awesome, that's what i really love in linters
> We have an `.env`
> 
> ```
>  LOGGER_LEVEL=info
> RAILS_ENV
> DB-NAME=testing
> DEbUG_hTTP=true
> DB_NAME=development
> ```
> 
> run `dotenv-linter --fix`
> 
> * remove leading spaces
> * replace incorrect delimeter to "_"
> * set '=' to empty keys
> * remove duplicated keys, sets value of the last key
> * order keys
> 
> We'll get:
> 
> ```
> DB_NAME=development
[...]

PR #202, opened 2020-05-24
Add an argument to show available checks (#199)

DDtKey:
Added functionality that displays a list of all checks if there is an argument --show-checks for #199
This argument is terminal and terminates the dotenv-linter after displaying a list of available check names.

The test was added for a new function, it may also make sense in the CLI test, but in this case, we will need to manually add names to the test in case of publication of new checks.

A usage example has been added to README.

#### ✔ Checklist:

- [X] Docs have been added / updated (for bug fixes / features).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### 🚀 Added
-- Add an argument to show available checks [#202](https://github.com/dotenv-linter/dotenv-linter/pull/202)  ([@DDtKey](https://github.com/DDtKey))
+- Add an argument to show available checks [#202](https://github.com/dotenv-linter/dotenv-linter/pull/202) ([@DDtKey](https://github.com/DDtKey))
 - Add the ability to skip checks [#178](https://github.com/dotenv-linter/dotenv-linter/pull/178) ([@mgrachev](https://github.com/mgrachev))
 - Add check: ExtraBlankLine [#180](https://github.com/dotenv-linter/dotenv-linter/pull/180) ([@evgeniy-r](https://github.com/evgeniy-r))
 - Add check: EndingBlankLine [#170](https://github.com/dotenv-linter/dotenv-linter/pull/170) ([@evgeniy-r](https://github.com/evgeniy-r))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### 🚀 Added
+- Add an argument to show available checks [#202](https://github.com/dotenv-linter/dotenv-linter/pull/202)  ([@DDtKey](https://github.com/DDtKey))
 - Add the ability to skip checks [#178](https://github.com/dotenv-linter/dotenv-linter/pull/178) ([@mgrachev](https://github.com/mgrachev))
 - Add check: ExtraBlankLine [#180](https://github.com/dotenv-linter/dotenv-linter/pull/180) ([@evgeniy-r](https://github.com/evgeniy-r))
 - Add check: EndingBlankLine [#170](https://github.com/dotenv-linter/dotenv-linter/pull/170) ([@evgeniy-r](https://github.com/evgeniy-r))
--- a/README.md
+++ b/README.md
@@ -177,6 +177,21 @@ $ dotenv-linter --skip UnorderedKey EndingBlankLine
 .env:2 DuplicatedKey: The FOO key is duplicated
 ```
 
+If you need to view all available checks, you can use the argument `--show-checks` (will be available in [v2.0.0](https://github.com/dotenv-linter/dotenv-linter/issues/199)):
+
+```shell script
+$ dotenv-linter --show-checks
+DuplicatedKey
+ExtraBlankLine
+IncorrectDelimiter
+LeadingCharacter
+KeyWithoutValue
+LowercaseKey
+QuoteCharacter
+SpaceCharacter
+UnorderedKey
+```
+
 ## ✅ Checks
 
 ### Duplicated Key
--- a/README.md
+++ b/README.md
@@ -182,10 +182,11 @@ If you need to view all available checks, you can use the argument `--show-check
 ```shell script
 $ dotenv-linter --show-checks
 DuplicatedKey
+EndingBlankLine
 ExtraBlankLine
 IncorrectDelimiter
-LeadingCharacter
 KeyWithoutValue
+LeadingCharacter
 LowercaseKey
 QuoteCharacter
 SpaceCharacter
--- a/src/checks.rs
+++ b/src/checks.rs
@@ -33,10 +33,13 @@ fn checklist() -> Vec<Box<dyn Check>> {
 }
 
 pub fn available_check_names() -> Vec<String> {
-    checklist()
+    let mut names: Vec<String> = checklist()
         .iter()
         .map(|check| check.name().to_string())
-        .collect()
+        .collect();
+    names.push(ending_blank_line::EndingBlankLineChecker::default().name().to_string());
+    names.sort();
+    names
 }
 
 pub fn run(lines: Vec<LineEntry>, skip_checks: &[&str]) -> Vec<Warning> {
@@ -189,5 +192,7 @@ mod tests {
             let check_name = check.name();
             assert!(available_check_names.contains(&check_name.to_string()));
         }
+        let ending_blank_line = ending_blank_line::EndingBlankLineChecker::default();
+        assert!(available_check_names.contains(&ending_blank_line.name().to_string()));
     }
 }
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -177,6 +177,21 @@ $ dotenv-linter --skip UnorderedKey EndingBlankLine
 .env:2 DuplicatedKey: The FOO key is duplicated
 ```
 
+If you need to view all available checks, you can use the argument `--show-checks` (will be available in [v2.0.0](https://github.com/dotenv-linter/dotenv-linter/issues/199)):
+
+```shell script
+$ dotenv-linter --show-checks
```
mstruebing, line 183: I wouldn't write that list out here, it gets outdated really fast if we would add new checks or remove something and to hold them in sync is easily forgotten.

What I would rather do is to write the checkname after each headline of the check i.e.

```
### Duplicated Key (`DuplicatedKey`)
```

or something.
What do you think @mgrachev ?

DDtKey: I think this is just an example and I need to understand that the result may vary from version to version.
That's why nowhere in the code did I start adding tests with a strict list, but README is just an example.

That is why I consider it important to add an `help` argument, as I described in [Ideas issue](https://github.com/dotenv-linter/dotenv-linter/issues/23#issuecomment-633460192)

README.md
```diff
@@ -177,6 +177,21 @@ $ dotenv-linter --skip UnorderedKey EndingBlankLine
...
+DuplicatedKey
+ExtraBlankLine
```
mgrachev, line 185: You missed `EndingBlankLine`.

src/checks.rs
```diff
@@ -32,6 +32,13 @@ fn checklist() -> Vec<Box<dyn Check>> {
     ]
 }
 
+pub fn available_check_names() -> Vec<String> {
+    checklist()
```
mgrachev, line 36: This method doesn't include the `EndingBlankLine` check.

src/checks.rs
```diff
@@ -174,4 +181,13 @@ mod tests {
 
         assert_eq!(expected, run(lines, &skip_checks));
     }
+
+    #[test]
+    fn check_name_list() {
+        let available_check_names = available_check_names();
+        for check in checklist() {
+            let check_name = check.name();
+            assert!(available_check_names.contains(&check_name.to_string()));
```
mgrachev, line 190: Please, add an integration test for that function as well, because right now it misses the `EndingBlankLine`check.

src/checks.rs
DDtKey: > This method doesn't include the `EndingBlankLine` check.

Yes indeed!
In this case, we still have to make a strict list of checks (as I wrote [above](https://github.com/dotenv-linter/dotenv-linter/pull/202#issue-422457368)), and when adding new ones, we will need to explicitly update the integration test

But, it seems to me, perhaps, it is conceptually more correct to add the attribute of the last line to the `LineEntry` and check directly inside the check, as in other cases.
Since we have an exception check that takes our logic out.
This, of course, is beyond the scope of this task, but this is an occasion to reflect.

What do you think about it?

src/checks.rs
DDtKey: > Please, add an integration test for that function as well, because right now it misses the `EndingBlankLine`check.

Yes indeed!
In this case, we still have to make a strict list of checks (as I wrote [above](https://github.com/dotenv-linter/dotenv-linter/pull/202#issue-422457368)), and when adding new ones, we will need to explicitly update the integration test

But, it seems to me, perhaps, it is conceptually more correct to add the attribute of the last line to the `LineEntry` and check directly inside the check, as in other cases.
Since we have an exception check that takes our logic out.
This, of course, is beyond the scope of this task, but this is an occasion to reflect.

What do you think about it?

DDtKey: @mgrachev 
And because it's specific check ("exception"), I doubt between integration test with hardcode strict check list and manually check in module test specific name (`EndingBlankLine`) as in the example below. 
Because it's stands out from the general checks API.

```suggestion
        let available_check_names = available_check_names();
        for check in checklist() {
            let check_name = check.name();
            assert!(available_check_names.contains(&check_name.to_string()));
        }
        let ending_blank_line = ending_blank_line::EndingBlankLineChecker::default();
        assert!(available_check_names.contains(&ending_blank_line.name().to_string()));
    }
```

mgrachev: > Yes indeed!
> In this case, we still have to make a strict list of checks (as I wrote [above](https://github.com/dotenv-linter/dotenv-linter/pull/202#issue-422457368)), and when adding new ones, we will need to explicitly update the integration test
> 
> But, it seems to me, perhaps, it is conceptually more correct to add the attribute of the last line to the `LineEntry` and check directly inside the check, as in other cases.
> Since we have an exception check that takes our logic out.
> This, of course, is beyond the scope of this task, but this is an occasion to reflect.
> 
> What do you think about it?

It's an interesting idea. Could you give some example code how you would implement it?

mgrachev: > @mgrachev
> And because it's specific check ("exception"), I doubt between integration test with hardcode strict check list and manually check in module test specific name (`EndingBlankLine`) as in the example below.
> Because it's stands out from the general checks API.

Okay, let's leave it like that.

DDtKey: > > Yes indeed!
> > In this case, we still have to make a strict list of checks (as I wrote [above](https://github.com/dotenv-linter/dotenv-linter/pull/202#issue-422457368)), and when adding new ones, we will need to explicitly update the integration test
> > But, it seems to me, perhaps, it is conceptually more correct to add the attribute of the last line to the `LineEntry` and check directly inside the check, as in other cases.
> > Since we have an exception check that takes our logic out.
> > This, of course, is beyond the scope of this task, but this is an occasion to reflect.
> > What do you think about it?
> 
> It's an interesting idea. Could you give some example code how would you implement it?

Yes, a little later I will give an example of what I mean

DDtKey: > It's an interesting idea. Could you give some example code how would you implement it?

For example, we can add total_number to `LineEntry`
```
pub struct LineEntry {
    pub number: usize,
    pub total_number: usize,
    pub file_path: PathBuf,
    pub raw_string: String,
}
```
And method to `impl LineEntry` :
```
pub fn is_last_line(&self) -> bool {
    self.total_number == self.number
}
```

And in the end we just can add check to `EndingBlankLineChecker`:
`if line.is_last_line() {...}`


It's have one issue: not persistent, if we add line to vector - need update all line entries.

In theory, dynamically adding strings to a vector is not possible, but a large volume of tests needs to be processed, since we do not use the constructor method for `LineEntry`.

Using a boolean flag instead of total_number will simplify the task, you only need to update the last element earlier.
But perhaps total_number may be of more value in terms of check api (can also check first line as example).

We can also calculate the number of lines at the very beginning, which will allow us to allocate a vector with initial capacity, as it:
```
let total_lines = reader.lines().count();
let mut lines: Vec<LineEntry> = Vec::with_capacity(total_lines);
```

DDtKey: @mgrachev But, this study is beyond the scope of this task and requires some refactoring.
Therefore, at the moment I am making changes, as talk above, and ask for a re-review of the code.
And this topic can be discussed in a separate issue.

Thanks for good feedback!

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### 🚀 Added
+- Add an argument to show available checks [#202](https://github.com/dotenv-linter/dotenv-linter/pull/202)  ([@DDtKey](https://github.com/DDtKey))
```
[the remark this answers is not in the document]
DDtKey: Updated in new commits, thanks!

src/lib.rs
```diff
@@ -65,6 +70,13 @@ pub fn run() -> Result<Vec<Warning>, Box<dyn Error>> {
         skip_checks = skip.collect();
     }
 
+    if args.is_present("show-checks") {
+        checks::available_check_names()
+            .iter()
+            .for_each(|name| println!("{}", name));
+        return Ok(vec![]);
```
mgrachev, line 77: Maybe instead of this we just exit with code 0?

src/checks.rs
mgrachev: I like that idea! 👏Thank you! Could you open an issue to implement it?

mgrachev: But I would like to move the `total_number` field to the `FileEntry` struct. It will be more correct.

DDtKey: Yes of course!  Soon I will create an issue, I also think this option(total_number) is more suitable

mgrachev: I have made some refactoring to implement that idea more easily.
https://github.com/dotenv-linter/dotenv-linter/pull/203

src/lib.rs
DDtKey: This approach seems really more transparent to terminal arguments.  Thank you! I will correct 😊

mgrachev:
@DDtKey Thanks a lot! You're awesome! ❤️

src/checks.rs
DDtKey: > I like that idea! :clap: Thank you! Could you open an issue to implement it?

Done #204

Issue #112, opened 2020-03-22
Provide more short way to install dotenv-linter


mgrachev:
Right now we can install `dotenv-linter` something like this:
```
# Linux
$ curl https://github.com/mgrachev/dotenv-linter/releases/download/v1.1.2/dotenv-linter-linux-x86_64.tar.gz -sSfL | tar -xzf - 

# Alpine Linux
$ wget https://github.com/mgrachev/dotenv-linter/releases/download/v1.1.2/dotenv-linter-alpine-x86_64.tar.gz -O - -q | tar -xzf -

# macOS
$ curl https://github.com/mgrachev/dotenv-linter/releases/download/v1.1.2/dotenv-linter-darwin-x86_64.tar.gz -sSfL | tar -xzf -
```

I would like to have a script which provides more short way to install it. For example:
```
curl -sfL https://raw.githubusercontent.com/reviewdog/reviewdog/master/install.sh | sh -s
```

Or something like that.

Macbet:
sounds like a great issues , I'll take it !

mgrachev:
@Macbet 👋Is there any progress with this issue?

DDtKey:
I'll try implement this.

PR #220, opened 2020-06-23
Provide more short way to install dotenv-linter

DDtKey:
Implemented a script for more short way to install dotenv-linter (#112).

The advantage of the solution is its cross-platform.
The script is executed with `curl`/`wget` and even _under Windows_ (checked on `MINGW`).

Complexity is the process of automatically declaring an environment variable. In some cases, this is only possible manually at the moment.

This is a working version, I also tested on Alpine Linux.

**But I would like to hear any ideas for improving or changing the approach.**

Perhaps when we come to a final decision, it makes sense to create a separate repository.
Although for installing binaries with a script, it looks suitable in this repo, so far it is not more than 1 file.

#### ✔ Checklist:

- [X] Docs have been added / updated (for bug fixes / features).

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,13 +59,13 @@ Articles about dotenv-linter:
 
 ```shell script
 # Linux / macOS / Windows (MINGW and etc)
-$ curl -sfl https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh | sh -s
+$ curl -sfL "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s
 
 # Alpine Linux (wget)
-$ wget -q -O - https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh | sh -s
+$ wget -q -O - "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s
 
 # Specific version
-$ curl -sfL https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh | sh -s "v1.2.0"
+$ curl -sfL "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s "v1.2.0"
 
 ```
 
--- a/README.md
+++ b/README.md
@@ -61,11 +61,11 @@ Articles about dotenv-linter:
 # Linux / macOS / Windows (MINGW and etc)
 $ curl -sSfL "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s
 
+# Specify installation directory and version.
+$ curl -sSfL "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s -- -b usr/local/bin v2.0.0
+
 # Alpine Linux (wget)
 $ wget -q -O - "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s
-
-# Specific version
-$ curl -sSfL "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s v1.2.0
 ```
 
 ### Homebrew / Linuxbrew
--- a/README.md
+++ b/README.md
@@ -58,14 +58,15 @@ Articles about dotenv-linter:
 ### Binary
 
 ```shell script
-# Linux
-$ curl https://github.com/dotenv-linter/dotenv-linter/releases/latest/download/dotenv-linter-linux-x86_64.tar.gz -sSfL | tar -xzf -
+# Linux / macOS / Windows (MINGW and etc)
+$ curl -sfl https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh | sh -s
 
-# Alpine Linux
-$ wget https://github.com/dotenv-linter/dotenv-linter/releases/latest/download/dotenv-linter-alpine-x86_64.tar.gz -O - -q | tar -xzf -
+# Alpine Linux (wget)
+$ wget -q -O - https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh | sh -s
 
-# macOS
-$ curl https://github.com/dotenv-linter/dotenv-linter/releases/latest/download/dotenv-linter-darwin-x86_64.tar.gz -sSfL | tar -xzf -
+# Specific version
+$ curl -sfL https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh | sh -s "v1.2.0"
+
 ```
 
 ### Homebrew / Linuxbrew
--- a/README.md
+++ b/README.md
@@ -58,14 +58,14 @@ Articles about dotenv-linter:
 ### Binary
 
 ```shell script
-# Linux / macOS / Windows (MINGW and etc)
-$ curl -sSfL "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s
+# Linux / macOS / Windows (MINGW and etc). Installs it into ./bin/ by default.
+$ curl -sSfL https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh | sh -s
 
 # Specify installation directory and version.
-$ curl -sSfL "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s -- -b usr/local/bin v2.0.0
+$ curl -sSfL https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh | sh -s -- -b usr/local/bin v2.0.0
 
 # Alpine Linux (wget)
-$ wget -q -O - "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s
+$ wget -q -O - https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh | sh -s
 ```
 
 ### Homebrew / Linuxbrew
--- a/README.md
+++ b/README.md
@@ -59,14 +59,13 @@ Articles about dotenv-linter:
 
 ```shell script
 # Linux / macOS / Windows (MINGW and etc)
-$ curl -sfL "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s
+$ curl -sSfL "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s
 
 # Alpine Linux (wget)
 $ wget -q -O - "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s
 
 # Specific version
-$ curl -sfL "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s "v1.2.0"
-
+$ curl -sSfL "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s v1.2.0
 ```
 
 ### Homebrew / Linuxbrew
--- a/docs/install.md
+++ b/docs/install.md
@@ -5,14 +5,14 @@ You can install the pre-compiled binary (in several different ways), use Docker 
 **Binary**
 
 ```shell script
-# Linux / macOS / Windows (MINGW and etc)
-$ curl -sSfL "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s
+# Linux / macOS / Windows (MINGW and etc). Installs it into ./bin/ by default.
+$ curl -sSfL https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh | sh -s
 
 # Specify installation directory and version.
-$ curl -sSfL "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s -- -b usr/local/bin v2.0.0
+$ curl -sSfL https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh | sh -s -- -b usr/local/bin v2.0.0
 
 # Alpine Linux (wget)
-$ wget -q -O - "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s
+$ wget -q -O - https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh | sh -s
 ```
 
 **Homebrew / Linuxbrew**
--- a/docs/install.md
+++ b/docs/install.md
@@ -5,14 +5,14 @@ You can install the pre-compiled binary (in several different ways), use Docker 
 **Binary**
 
 ```shell script
-# Linux
-$ curl https://github.com/dotenv-linter/dotenv-linter/releases/latest/download/dotenv-linter-linux-x86_64.tar.gz -sSfL | tar -xzf -
+# Linux / macOS / Windows (MINGW and etc)
+$ curl -sSfL "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s
 
-# Alpine Linux
-$ wget https://github.com/dotenv-linter/dotenv-linter/releases/latest/download/dotenv-linter-alpine-x86_64.tar.gz -O - -q | tar -xzf -
+# Alpine Linux (wget)
+$ wget -q -O - "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s
 
-# macOS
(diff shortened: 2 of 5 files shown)


DDtKey:
One option is to install `dotenv-linter` at the system level(`/opt/dotenv-linter` for example) and create a symlink in `/usr/local/bin`.  
This will work without any export of variables at different shell levels.

But it will require the user rights to `sudo`.  It will not be possible for different users to refer to different versions, and, unfortunately, such a scenario will have many exceptions for Windows :thinking:

mstruebing:
Very nice work, thank you, unfortunately I'm a bit low on time right now. But I will have a detailed look soonish, at latest this weekend. :rocket:

mgrachev:
@DDtKey That's awesome! Thank you! 🔥

DDtKey:
@dotenv-linter/maintainers glad to hear any comments and discuss this

README.md
```diff
@@ -58,14 +58,15 @@ Articles about dotenv-linter:
 ### Binary
 
 ```shell script
-# Linux
-$ curl https://github.com/dotenv-linter/dotenv-linter/releases/latest/download/dotenv-linter-linux-x86_64.tar.gz -sSfL | tar -xzf -
+# Linux / macOS / Windows (MINGW and etc)
+$ curl -sfL "https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh" | sh -s
```
mgrachev, line 62:
```suggestion
$ curl -sSfL https://raw.githubusercontent.com/dotenv-linter/dotenv-linter/master/install.sh | sh -s
```
I'd like to see an error if something went wrong

install.sh
```diff
@@ -0,0 +1,211 @@
+#!/bin/bash
+
+set -u
```
mgrachev, line 3: What does it mean?

install.sh
```diff
@@ -0,0 +1,211 @@
...
+
+main() {
+    require_cmd uname
+    require_cmd mkdir
+    require_cmd grep
+    require_cmd cut
+    require_cmd rm
+
+    local _target_version
+    if [ $# -eq 0 ]; then
```
mgrachev, line 18: Can you tell me more about that?

install.sh
```diff
@@ -0,0 +1,211 @@
...
+    case $_arch in
+        win-*) _ext=".zip" ;;
+        *) _ext=".tar.gz" ;;
+    esac
+
+    local _archive_name="${DOTENV_LINTER}-${_arch}${_ext}"
+    local _url="${DOTENV_LINTER_GITHUB}/releases/download/${_target_version}/${_archive_name}"
+
+    # Installation
+    local _dir="$HOME/.${DOTENV_LINTER}"
```
mgrachev, line 38: I'd like to install dotenv-linter to the `bin` directory in the current directory.
Mostly, this script will be used in CI-services. So, I wouldn't like to install the project globally.

I've seen a lot of projects with the similar approach. For example:
https://github.com/reviewdog/reviewdog#installation
https://github.com/reviewdog/reviewdog/blob/master/install.sh#L12

install.sh
```diff
@@ -0,0 +1,211 @@
...
+    local _version_dir="${_dir}/${_target_version}"
+
+    if [ ! -e "$_version_dir/${DOTENV_LINTER}" ]; then
+        mkdir -p "$_version_dir"
+
+        local _archive_path="${_version_dir}/${_archive_name}"
+        download "${_url}" "${_archive_path}" || return 1
```
mgrachev, line 45: Maybe we will download an archive into temp directory?

install.sh
```diff
@@ -0,0 +1,211 @@
...
+            *.zip)
+                required_cmd unzip
+                unzip "${_archive_path}" -d "${_version_dir}"
+            ;;
+            *)  require_cmd tar
+                tar -xzf "${_archive_path}" -C "${_version_dir}"
+            ;;
+        esac
+
+        rm "${_archive_path}"
```
mgrachev, line 57: I'd like to remove only temp directory which is inside `/tmp`.

install.sh
```diff
@@ -0,0 +1,211 @@
...
+
+# $1 - dotenv-linter dir, $2 - target version
+setup_executable() {
+    local _bin_dir="$1/bin"
+    if [ ! -d "$_bin_dir" ]; then
+	    mkdir -p "$_bin_dir"
+    fi
+
+    local _bin_file="${_bin_dir}/${DOTENV_LINTER}"
+    ln -sf "$1/$2/${DOTENV_LINTER}" "${_bin_file}"
```
mgrachev, line 79: Why do we do that?

install.sh
```diff
@@ -0,0 +1,211 @@
...
+    local _shell_profile
+    case ${SHELL-undefined} in
+        /bin/zsh) _shell_profile=~/".zshrc" ;;
+        *) _shell_profile=~/".bashrc" ;;
+	esac
+
+    if [ -e "$_shell_profile" ]; then
+        case :$PATH: in
+            *:$_bin_dir:*) ;; # do nothing, already exists
+            *)  echo "export PATH=\"$_bin_dir:\$PATH\"" >> "${_shell_profile}"
```
mgrachev, line 90: We shouldn't do that. To tell the truth, I hate scripts which modify files in my local machine.
Instead of this we can just print this command. If a user wants, he will do it himself.

install.sh
```diff
@@ -0,0 +1,211 @@
...
+                println "Will be available in new terminal sessions or after update config: \". ${_shell_profile}\""
+            ;;
+        esac
+    fi
+
+    if ! cmd_exists dotenv-linter ; then
+        println ""
+	    println "Manually add the directory to your environment:"
+        println "   export PATH=\"\$PATH:${_bin_dir}\""
+        println ""
```
mgrachev, line 101:
```suggestion
        println ""
        println "Manually add the directory to your environment:"
        println "   export PATH=\"\$PATH:${_bin_dir}\""
        println ""
```

install.sh
```diff
@@ -0,0 +1,211 @@
...
+    local _ostype _cputype _clibtype
+    _ostype="$(uname -s)"
+    _cputype="$(uname -m)"
+    _clibtype="gnu"
+
+    if [ "$_cputype" != "x86_64" ]; then
+	    err "Error: Unsupported architecture $_cputype. Only x86_64 binaries are available."
+    fi
+
+    if [ "$_ostype" = Linux ]; then
```
mgrachev, line 127:
```suggestion
    if [ "$_ostype" = "Linux" ]; then
```
?

install.sh
```diff
@@ -0,0 +1,211 @@
...
+    command -v "$1" > /dev/null 2>&1
+}
+
+err() {
+    println "$1" >&2
+    exit 1
+}
+
+println() {
+    printf '%s installer:   %s\n' "${DOTENV_LINTER}" "$1"
```
mgrachev, line 208:
```suggestion
    printf '%s installer: %s\n' "${DOTENV_LINTER}" "$1"
```

mgrachev:
@DDtKey I've left some comments. Please, look at them.

Also, please look at this example of installing via a script: https://github.com/reviewdog/reviewdog#installation.
I like that API. Can we have something like that? 🙂

mgrachev:
Also, it would be nice if we could test this script via GitHub Actions on Linux, Mac and Windows.
What do you think about it?

install.sh
DDtKey: Good catch!)

install.sh
DDtKey: Most of all, I wanted to discuss this particular approach, for some reason it seemed to me that the goal was precisely in a fully automatic installation in the environment.

But we will do exactly as you say, because I like this approach more, too

install.sh
DDtKey: I was thinking about supporting multi-versioning, but that seems overkill.
Symbolic links are used by many programs for local use without the need to replace the entire versioned package.

We will change this, because I will adhere to the approach mentioned above.

install.sh
DDtKey: Yes, I agree :+1:

install.sh
DDtKey: Actually, I agree.
I posted this version to consult exactly on the installation method.

By and large, I am inspired by [rustup](sh.rustup.rs) _sh installer_. It has a similar way.
I also looked at the reviewdog installer, it looks good, and I also wanted to use a special flag for `bindir` and will do it soon, since I like the approach without changing the environment variables more

install.sh
DDtKey: What exactly are you interested in?
`$#` - the number of arguments, if a specific version is not specified, we get the last by means of API github.

Without the need for a multi-version installation, I will also simplify downloading the latest version (remove the extra github api call)

install.sh
DDtKey: Treat unset variables as an error when substituting. Maybe `set -e` suits us better

install.sh
```diff
@@ -0,0 +1,211 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
DDtKey: Yes, from the syntax bash, I used only local variables, this is pretty painless

DDtKey:
> Also, it would be nice if we could test this script via GitHub Actions on Linux, Mac and Windows.
> What do you think about it?

Good idea, but until release 2.1.0 we cannot run a script test for Windows, because there are no published versions.
He will fall.
We can certainly expect this now, and after the release update.
And perhaps it’s worth considering as a separate issue :thinking:

install.sh
mgrachev: I think so too 🙂

install.sh
mgrachev: Maybe we don't need to have it in our script, because we should do this script as simple as possible.
We can add to README an instruction how to install on a local machine and add this line too.
What dot you think?

mgrachev:
> Good idea, but until release 2.1.0 we cannot run a script test for Windows, because there are no published versions.
> He will fall.
> We can certainly expect this now, and after the release update test behavior 🤔

Agree. In the first iteration we can add tests only for Mac and Linux.

DDtKey:
> Also, it would be nice if we could test this script via GitHub Actions on Linux, Mac and Windows.
> What do you think about it?

I made the changes we talked about, also added CI tests

README.md
```diff
@@ -58,14 +58,14 @@ Articles about dotenv-linter:
 ### Binary
 
 ```shell script
-# Linux
-$ curl https://github.com/dotenv-linter/dotenv-linter/releases/latest/download/dotenv-linter-linux-x86_64.tar.gz -sSfL | tar -xzf -
+# Linux / macOS / Windows (MINGW and etc)
```
mgrachev, line 61:
```suggestion
# Linux / macOS / Windows (MINGW and etc). Installs it into ./bin/ by default.
```

docs/install.md
```diff
@@ -5,14 +5,14 @@ You can install the pre-compiled binary (in several different ways), use Docker
 **Binary**
 
 ```shell script
-# Linux
-$ curl https://github.com/dotenv-linter/dotenv-linter/releases/latest/download/dotenv-linter-linux-x86_64.tar.gz -sSfL | tar -xzf -
+# Linux / macOS / Windows (MINGW and etc)
```
mgrachev, line 8:
```suggestion
# Linux / macOS / Windows (MINGW and etc). Installs it into ./bin/ by default.
```

install.sh
```diff
@@ -0,0 +1,188 @@
...
+$this: download binaries for dotenv-linter
+
+USAGE:
+    $this [FLAGS] [OPTIONS] <tag>
+
+FLAGS:
+    -h, --help      Prints help information
+
+OPTIONS:
+    -b, --bindir <DIR_PATH>     Sets bindir or installation directory, Defaults to ./bin
```
mgrachev, line 22:
```suggestion
    -b, --bindir <DIR_PATH>     Sets bindir or installation directory. Defaults to ./bin
```

install.sh
```diff
@@ -0,0 +1,188 @@
...
+    exit 2
+}
+
+parse_args() {
+  BINDIR=${BINDIR:-./bin}
+  while [ "$#" -gt 0 ]; do
+    case $1 in
+        -h|--help)
+            usage "$0"
+            shift #past argument
```
mgrachev, line 36:
```suggestion
            shift # past argument
```

install.sh
```diff
@@ -0,0 +1,188 @@
...
+
+    case $_arch in
+        win-*) _ext=".zip" ;;
+        *) _ext=".tar.gz" ;;
+    esac
+
+    _archive_name="${DOTENV_LINTER}-${_arch}${_ext}"
+
+    if [ -z "${TAG}" ]; then
+        println "The latest version will be installed."
```
mgrachev, line 70: I'd like to know which version will be installed 🙂

DDtKey: > I'd like to know which version will be installed 🙂

In that case, later, I will return the function of getting the latest version using the github API :)

I deleted it just to reduce the number of requests, to one

mgrachev:
@mstruebing What do you think about it?

DDtKey:
I just resolved the conflict in `CHANGELOG.md`.

But faced with an unpleasant feature, when we use _api github_, it has limitations on the number of calls.

We need this to get the "latest" version.
I think this is not a problem in general, but CI starts quite often and this time it ran into a limit, which could happen again.

I see three possible solutions:
1) In CI tests use a specific version to test the script (but we do not test to get the latest version unfortunately)
2) As it was before: do not display the latest version number (i.e., use a static link to access the github, bypassing the api).
3) Modify the script, for example, in case of an error in obtaining the version in api, download the latest version directly, bypassing the api

What do you think? 
@dotenv-linter/core

install.sh
```diff
@@ -0,0 +1,212 @@
...
+
+get_latest_version() {
+    require_cmd cut
+
+    download "${DOTENV_LINTER_RELEASE_API}" || return 1
+    _latest_version=$(
+         echo "$RETVAL" | grep '"tag_name":' | cut -d'"' -f4
+    )
+
+    RETVAL=${_latest_version}
```
mstruebing, line 116: It seems like `RETVAL` is a global variable, I've never seen this pattern until now but it seems usable.
Personally I like the `echo` way a bit better, as it eliminates the usage of global variables.

You can read about it here in example 2: https://linuxhint.com/return-string-bash-functions/

Anyway - if you want to stick to the `RETVAL` way, go for it. :)

mstruebing:
> 1. In CI tests use a specific version to test the script (but we do not test to get the latest version unfortunately)
In tests we could use https://github.com/dotenv-linter/dotenv-linter/releases/latest/download/dotenv-linter-os-arch.gz to always have the newest version

Why do we need to print the version number on install anyway? Wouldn't it be enough to say after installation finishes we present the user the version (by using `dotenv-linter --version`)?

install.sh
DDtKey: Actually, I liked this approach in sh scripts.
I met him a couple of times, for example in a [rustup instillation script](https://sh.rustup.rs/) :relaxed:

mstruebing: Okay, let's leave it like this :) It is just a personal opinion and I don't think this will be the reason why our install script will break at some point.

DDtKey:
> > 1. In CI tests use a specific version to test the script (but we do not test to get the latest version unfortunately)
> 
> In tests we could use https://github.com/dotenv-linter/dotenv-linter/releases/latest/download/dotenv-linter-os-arch.gz to always have the newest version
> 
> Why do we need to print the version number on install anyway? Wouldn't it be enough to say after installation finishes we present the user the version (by using `dotenv-linter --version`)?

This approach was used before [discussing the print of the installed version](https://github.com/dotenv-linter/dotenv-linter/pull/220#discussion_r449756002)

But in fact, I agree, this option will solve the problems of limiting api github and reduce the number of calls to `curl` / `wget` :thinking: 
What do you think? @mgrachev

mgrachev:
> Why do we need to print the version number on install anyway? Wouldn't it be enough to say after installation finishes we present the user the version (by using `dotenv-linter --version`)?

I like this idea 👍

mstruebing:
> This approach was used before discussing the print of the installed version

Sorry, I didn't followed the conversation :)

DDtKey:
@dotenv-linter/core thanks for feedback. 
I returned using the latest version static URL to minimize API calls. User can always use `dotenv-linter -v`

install.sh
```diff
@@ -0,0 +1,188 @@
...
+        *)  require_cmd tar
+            tar -xzf "${_archive_path}" -C "${_temp_dir}"
+            _file_name="${DOTENV_LINTER}"
+        ;;
+    esac
+
+    test ! -d "${BINDIR}" && install -d "${BINDIR}"
+
+    install "${_temp_dir}/${_file_name}" "${BINDIR}/"
+    println "Successfully installed to ${BINDIR}/${_file_name}"
```
mgrachev, line 98: It would be nice if we got the version of dotenv-linter and printed it here.
What do you think?

mgrachev:
> I returned using the latest version static URL to minimize API calls. User can always use `dotenv-linter -v`

I thought that @mstruebing suggested getting the version of dotenv-linter using the `dotenv-linter -v` command and printing it after installation is completed.

DDtKey:
> > I returned using the latest version static URL to minimize API calls. User can always use `dotenv-linter -v`
> 
> I thought that @mstruebing suggested getting the version of dotenv-linter using the `dotenv-linter -v` command and printing it after installation is completed. Or am I misunderstood?

This is my misunderstanding, I apologize, it option really impresses me. Thank you :+1:

PR #221, opened 2020-06-24
Consider blank lines in UnorderedKey check

mgrachev:
In some projects, I work with large `.env` files which have more than 200 lines.
Sorting these files is a really big problem.
The solution is to use groups of lines separated by blank lines. For example:
```env
❌ Wrong
FOO=BAR
BAR=FOO

✅ Correct
BAR=FOO
FOO=BAR

✅ Correct
FOO=BAR

BAR=FOO
```

#### ✔ Checklist:

### DIFF
--- a/src/checks/unordered_key.rs
+++ b/src/checks/unordered_key.rs
@@ -31,6 +31,11 @@ impl Default for UnorderedKeyChecker<'_> {
 
 impl Check for UnorderedKeyChecker<'_> {
     fn run(&mut self, line: &LineEntry) -> Option<Warning> {
+        if line.is_empty() {
+            self.keys.clear();
+            return None;
+        }
+
         let key = line.get_key()?;
         self.keys.push(key.clone());
         let mut sorted_keys = self.keys.clone();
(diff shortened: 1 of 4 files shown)


src/checks/unordered_key.rs
```diff
@@ -369,4 +374,48 @@ mod tests {
...
+                },
+                None,
+            ),
+            (
+                LineEntry {
+                    number: 2,
+                    file: FileEntry {
+                        path: PathBuf::from(".env"),
+                        file_name: ".env".to_string(),
+                        total_lines: 2,
```
mstruebing, line 411: Shouldnt that be three?

```suggestion
                        total_lines: 3,
```

src/checks/unordered_key.rs
```diff
@@ -369,4 +374,48 @@ mod tests {
...
+                    },
+                    raw_string: String::from("FOO=BAR"),
+                },
+                None,
+            ),
+            (
+                LineEntry {
+                    number: 1,
```
mstruebing, line 395: Shouldnt that be two?

```suggestion
                    number: 2,
```

src/checks/unordered_key.rs
```diff
@@ -369,4 +374,48 @@ mod tests {
...
+                        file_name: ".env".to_string(),
+                        total_lines: 2,
+                    },
+                    raw_string: String::from(""),
+                },
+                None,
+            ),
+            (
+                LineEntry {
+                    number: 2,
```
mstruebing, line 407: Shouldn't that be three?

```suggestion
                    number: 3,
```

DDtKey:
> Making it configurable is a good thing I think, but as you've stated I also think that this should be the default behaviour.
> Configurability can also be handled in a separate PR to have small PRs which makes the life of everyone easier :)

Yes, of course, I am fully in favor of this being a separate PR.

mgrachev:
@DDtKey @mstruebing I'm not sure 

Configurability is a good idea, but I'm not sure that we need it in this check.
Right now, we have very simple checks that are not configurable and implementing this for one check will add extra complexity.

I would like to return to this discussion in the future if we need to implement configurability for several checks.
What do you think?

PR #227, opened 2020-06-29
Move show-checks flag to main.rs

mgrachev:
What have been done:
- [x] Sorting is removed from the `available_check_names` method, because all checks are already in alphabetical order;
- [x] The `show checks` flag is moved to `main.rs` file, because it allowed to exit from the program.

#### ✔ Checklist:

- [x] This PR has been added to [CHANGELOG.md](https://github.com/dotenv-linter/dotenv-linter/blob/master/CHANGELOG.md) (at the top of the list).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,74 +1,19 @@
 use crate::common::*;
 
-use clap::Arg;
 use std::error::Error;
-use std::ffi::OsStr;
 use std::path::PathBuf;
-use std::{env, process};
 
 mod checks;
 mod common;
 mod fs_utils;
 
-fn get_args(current_dir: &OsStr) -> clap::ArgMatches {
-    clap::App::new(env!("CARGO_PKG_NAME"))
-        .about(env!("CARGO_PKG_DESCRIPTION"))
-        .author(env!("CARGO_PKG_AUTHORS"))
-        .version(env!("CARGO_PKG_VERSION"))
-        .version_short("v")
-        .arg(
-            Arg::with_name("input")
-                .help("files or paths")
-                .index(1)
-                .default_value_os(current_dir)
-                .required(true)
-                .multiple(true),
-        )
-        .arg(
-            Arg::with_name("exclude")
-                .short("e")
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -1,75 +1,22 @@
...
-                .multiple(true)
-                .takes_value(true),
-        )
-        .arg(
-            Arg::with_name("show-checks")
-                .long("show-checks")
-                .help("Shows list of available checks"),
-        )
-        .get_matches()
+pub fn show_checks() -> Vec<String> {
```
DDtKey, line 10: I suggest use: 
```suggestion
pub use checks::available_check_names;
```
Because the name of the function(`show_checks`) indicates the action, but in fact it’s just a delegate.

Or make this function with `println`, but better stay all prints in `main.rs`

src/lib.rs
```diff
@@ -1,75 +1,22 @@
...
+    checks::available_check_names()
```
DDtKey, line 11:
```suggestion
```

src/main.rs
```diff
@@ -1,21 +1,64 @@
...
+    let args = get_args(current_dir.as_os_str());
 
-fn main() {
-    match dotenv_linter::run() {
-        Ok(warnings) => {
-            if warnings.is_empty() {
-                process::exit(0);
-            }
+    if args.is_present("show-checks") {
+        dotenv_linter::show_checks()
```
DDtKey, line 11:
```suggestion
        dotenv_linter::available_check_names()
```

src/lib.rs
```diff
@@ -1,75 +1,22 @@
...
 }
```
DDtKey, line 12:
```suggestion
```

DDtKey:
It’s a good idea to leave terminal operations in main.rs! :rocket: 
Sorry for some mess with suggestions for improvement.

src/lib.rs
mgrachev: Good catch 🙂

PR #228, opened 2020-06-30
Autofix

evgeniy-r:
I have implemented automatic correction of the contents of files based on the results of the check.
When displaying, all warnings are marked as Fixed or Unfixed.
If all warnings are fixed, exit code is 0.

Just one kind of checks is fixable for now (LowercaseKey), but I have create the scaffold for implementing all kinds of checks.

The CLI key for the autofix mode is `-f`/`--fix`.

A few comments about the design:

The separate trait `Fix` was made for fixes (I tried to keep some symmetry with `checks`), but it is possible to expand the trait `Check` itself with this functions too.

There are two groups of fixes: some can be implemented by working only with the line that caused warning, for others you need to work with the entire line collection.

The `Fix` trait allows you to work with both groups of fixes due to default function implementations.

#### ✔ Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### 🚀 Added
+- Add autofix feature (just one check for now - LowercaseKey) [#228](https://github.com/dotenv-linter/dotenv-linter/pull/228) ([@evgeniy-r](https://github.com/evgeniy-r))
 - Add installation CI test for Windows (via `install.sh`) [#235](https://github.com/dotenv-linter/dotenv-linter/pull/235) ([@DDtKey](https://github.com/DDtKey))
 
 ### 🔧 Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### 🚀 Added
-- Add autofix feature (just one check for now - LowercaseKey) [#228](https://github.com/dotenv-linter/dotenv-linter/pull/228) ([@evgeniy-r](https://github.com/evgeniy-r))
+- Add autofix feature (for LowercaseKey check) [#228](https://github.com/dotenv-linter/dotenv-linter/pull/228) ([@evgeniy-r](https://github.com/evgeniy-r))
 - Add installation CI test for Windows (via `install.sh`) [#235](https://github.com/dotenv-linter/dotenv-linter/pull/235) ([@DDtKey](https://github.com/DDtKey))
 
 ### 🔧 Changed
--- a/README.md
+++ b/README.md
@@ -209,6 +209,18 @@ TrailingWhitespace
 UnorderedKey
 ```
 
+`dotenv-linter` can also automatically fix warnings in the files. Currently only one kind of warnings is fixed
+(`LowercaseKey`). You should use the argument `--fix` (or its short version `-f`) for this.
+
+```shell script
+$ dotenv-linter -f
+Fixed warnings:
+.env:3 LowercaseKey: The foo key should be in uppercase
+
+Unfixed warnings:
+.env:2 DuplicatedKey: The BAR key is duplicated
+```
+
 ## ✅ Checks
 
 ### Duplicated Key
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ UnorderedKey
 ```
 
 `dotenv-linter` can also automatically fix warnings in the files. Currently only one kind of warnings is fixed
-(`LowercaseKey`). You should use the argument `--fix` (or its short version `-f`) for this.
+(`LowercaseKey`). You should use the argument `--fix` (or its short version `-f`) for this (will be available in [v2.2.0](https://github.com/dotenv-linter/dotenv-linter/issues/238):
 
 ```shell script
 $ dotenv-linter -f
--- a/src/common.rs
+++ b/src/common.rs
@@ -184,7 +184,6 @@ mod tests {
 
         mod from {
             use super::*;
-            use std::fs::remove_file;
 
             #[test]
             fn path_without_file_test() {
@@ -203,7 +202,7 @@ mod tests {
                 assert_eq!(
                     Some((
                         FileEntry {
-                            path: path.clone(),
+                            path,
                             file_name,
                             total_lines: 0
                         },
@@ -211,7 +210,7 @@ mod tests {
                     )),
                     f
                 );
-                remove_file(path).expect("temp file deleted");
+                dir.close().expect("temp dir deleted");
             }
         }
 
--- a/src/common.rs
+++ b/src/common.rs
@@ -28,6 +28,10 @@ impl Warning {
     pub fn mark_as_fixed(&mut self) {
         self.is_fixed = true;
     }
+
+    pub fn mark_as_unfixed(&mut self) {
+        self.is_fixed = false;
+    }
 }
 
 impl fmt::Display for Warning {
--- a/src/common.rs
+++ b/src/common.rs
@@ -4,9 +4,10 @@ use std::path::PathBuf;
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct Warning {
-    check_name: String,
+    pub check_name: String,
     line: LineEntry,
     message: String,
+    pub is_fixed: bool,
 }
 
 impl Warning {
@@ -16,7 +17,16 @@ impl Warning {
             line,
             check_name,
             message,
+            is_fixed: false,
         }
+    }
+
+    pub fn line_number(&self) -> usize {
+        self.line.number
+    }
+
+    pub fn mark_as_fixed(&mut self) {
+        self.is_fixed = true;
     }
 }
 
@@ -174,7 +184,6 @@ mod tests {
 
         mod from {
             use super::*;
-            use std::env::temp_dir;
             use std::fs::remove_file;
 
             #[test]
@@ -186,7 +195,8 @@ mod tests {
             #[test]
             fn path_with_file_test() {
                 let file_name = String::from(".env");
-                let path = temp_dir().join(&file_name);
+                let dir = tempfile::tempdir().expect("create temp dir");
+                let path = dir.path().join(&file_name);
                 fs::File::create(&path).expect("create testfile");
 
                 let f = FileEntry::from(path.clone());
--- a/src/fixes.rs
+++ b/src/fixes.rs
@@ -12,7 +12,7 @@ trait Fix {
     ) -> Option<usize> {
         let mut count: usize = 0;
         for warning in warnings {
-            let line = &mut lines[warning.line_number() - 1];
+            let line = lines.get_mut(warning.line_number() - 1)?;
             if self.fix_line(line).is_some() {
                 warning.mark_as_fixed();
                 count += 1;
@@ -52,8 +52,17 @@ pub fn run(warnings: &mut [Warning], lines: &mut Vec<LineEntry>) -> usize {
             .iter_mut()
             .filter(|w| w.check_name == fixer.name())
             .collect();
+
         if !fixer_warnings.is_empty() {
-            count += fixer.fix_warnings(fixer_warnings, lines).unwrap_or(0);
+            match fixer.fix_warnings(fixer_warnings, lines) {
+                Some(fixer_count) => count += fixer_count,
+                None => {
+                    for warning in warnings {
+                        warning.mark_as_unfixed();
+                    }
+                    return 0;
+                }
+            }
         }
     }
 
@@ -127,6 +136,30 @@ mod tests {
             "KeyWithoutValue",
             String::from("The C key should be with a value or have an equal sign"),
         )];
+
+        assert_eq!(0, run(&mut warnings, &mut lines));
+        assert!(!warnings[0].is_fixed);
+    }
+
+    #[test]
+    fn run_when_lines_do_not_fit_numbers_test() {
+        let mut lines = vec![
+            line_entry(1, 3, "a=B"),
+            line_entry(4, 3, "c=D"),
+            blank_line_entry(3, 3),
+        ];
+        let mut warnings = vec![
+            Warning::new(
+                lines[0].clone(),
+                "LowercaseKey",
+                String::from("The a key should be in uppercase"),
+            ),
+            Warning::new(
+                lines[1].clone(),
+                "LowercaseKey",
+                String::from("The c key should be in uppercase"),
+            ),
+        ];
 
         assert_eq!(0, run(&mut warnings, &mut lines));
         assert!(!warnings[0].is_fixed);
--- a/src/fixes.rs
+++ b/src/fixes.rs
@@ -1,0 +1,134 @@
+use crate::common::*;
+
+mod lowercase_key;
+
+trait Fix {
+    fn name(&self) -> &str;
+
+    fn fix_warnings(
+        &self,
+        warnings: Vec<&mut Warning>,
+        lines: &mut Vec<LineEntry>,
+    ) -> Option<usize> {
+        let mut count: usize = 0;
+        for warning in warnings {
+            let line = &mut lines[warning.line_number() - 1];
+            if self.fix_line(line).is_some() {
+                warning.mark_as_fixed();
+                count += 1;
+            }
+        }
+
+        Some(count)
+    }
+
+    fn fix_line(&self, _line: &mut LineEntry) -> Option<()> {
+        None
+    }
+}
+
+// TODO: skip fixes (like checks)
+// The fix order is matter
+fn fixlist() -> Vec<Box<dyn Fix>> {
+    vec![
+        // At first we run the fixers that handle a single line entry (they use default
+        // implementation of the fix_warnings() function)
+        Box::new(lowercase_key::LowercaseKeyFixer::default()),
+        // Then we should run the fixers that handle the line entry collection at whole.
+        // And at the end we should run the fixer for ExtraBlankLine check (because the previous
+        // fixers can create additional extra blank lines).
+    ]
+}
+
+pub fn run(warnings: &mut [Warning], lines: &mut Vec<LineEntry>) -> usize {
+    if warnings.is_empty() {
+        return 0;
+    }
+
+    let mut count = 0;
+    for fixer in fixlist() {
+        // We can optimize it: create check_name:warnings map in advance
+        let fixer_warnings: Vec<&mut Warning> = warnings
+            .iter_mut()
+            .filter(|w| w.check_name == fixer.name())
+            .collect();
+        if !fixer_warnings.is_empty() {
+            count += fixer.fix_warnings(fixer_warnings, lines).unwrap_or(0);
+        }
+    }
+
+    count
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::path::PathBuf;
+
+    fn line_entry(number: usize, total_lines: usize, str: &str) -> LineEntry {
+        LineEntry {
+            number,
+            file: FileEntry {
+                path: PathBuf::from(".env"),
+                file_name: ".env".to_string(),
+                total_lines,
+            },
+            raw_string: String::from(str),
+        }
+    }
+
+    fn blank_line_entry(number: usize, total_lines: usize) -> LineEntry {
+        LineEntry {
+            number,
+            file: FileEntry {
+                path: PathBuf::from(".env"),
+                file_name: ".env".to_string(),
+                total_lines,
+            },
+            raw_string: String::from("\n"),
+        }
+    }
+
+    #[test]
+    fn run_with_empty_warnings_test() {
+        let mut lines = vec![line_entry(1, 2, "A=B"), blank_line_entry(2, 2)];
+        let mut warnings: Vec<Warning> = Vec::new();
+
+        assert_eq!(0, run(&mut warnings, &mut lines));
+    }
+
+    #[test]
+    fn run_with_fixable_warning_test() {
+        let mut lines = vec![
+            line_entry(1, 3, "A=B"),
+            line_entry(2, 3, "c=d"),
+            blank_line_entry(3, 3),
+        ];
+        let mut warnings = vec![Warning::new(
+            lines[1].clone(),
+            "LowercaseKey",
+            String::from("The c key should be in uppercase"),
+        )];
+
+        assert_eq!(1, run(&mut warnings, &mut lines));
+        assert_eq!("C=d", lines[1].raw_string);
+        assert!(warnings[0].is_fixed);
+    }
+
+    #[test]
+    fn run_with_unfixable_warning_test() {
+        let mut lines = vec![
+            line_entry(1, 3, "A=B"),
+            line_entry(2, 3, "C"),
+            blank_line_entry(3, 3),
+        ];
+        let mut warnings = vec![Warning::new(
+            lines[1].clone(),
+            "KeyWithoutValue",
+            String::from("The C key should be with a value or have an equal sign"),
+        )];
+
+        assert_eq!(0, run(&mut warnings, &mut lines));
+        assert!(!warnings[0].is_fixed);
+    }
+}
--- a/src/fs_utils.rs
+++ b/src/fs_utils.rs
@@ -51,7 +51,7 @@ pub fn write_file(path: &PathBuf, lines: Vec<LineEntry>) -> io::Result<()> {
 mod tests {
     use super::*;
     use crate::common::FileEntry;
-    use std::fs::{self, remove_file, File};
+    use std::fs::{self, File};
 
     fn run_relative_path_asserts(assertions: Vec<(&str, &str, &str)>) {
         for (target, base, relative) in assertions {
@@ -102,7 +102,7 @@ mod tests {
         // The result of `fs_utils` must match `std::fs` for non-Windows platform
         assert_eq!(canonical_path, fs_canonical_path);
 
-        remove_file(path).expect("temp file deleted");
+        dir.close().expect("temp dir deleted");
     }
 
     #[test]
@@ -127,7 +127,7 @@ mod tests {
         assert_eq!(canonical_path, dunce_canonical_path);
         assert!(!contains_unc);
 
-        remove_file(path).expect("temp file deleted");
+        dir.close().expect("temp dir deleted");
     }
 
     #[test]
@@ -166,6 +166,6 @@ mod tests {
             fs::read(path.as_path()).expect("file read").as_slice()
         );
 
-        remove_file(path).expect("temp file deleted");
+        dir.close().expect("temp dir deleted");
     }
 }
--- a/src/fs_utils.rs
+++ b/src/fs_utils.rs
@@ -1,3 +1,6 @@
+use crate::common::LineEntry;
+use std::fs::File;
+use std::io::{self, Write};
 use std::path::PathBuf;
 
 /// For the Windows platform, we need to remove the UNC prefix.
@@ -31,11 +34,24 @@ pub fn get_relative_path(target_path: &PathBuf, base_path: &PathBuf) -> Option<P
     Some(relative_path)
 }
 
+/// In the future versions we should create a backup copy, or at least notify the user about it
+pub fn write_file(path: &PathBuf, lines: Vec<LineEntry>) -> io::Result<()> {
+    let mut file = File::create(path)?;
+
+    // We don't write the last line, because it contains only LF (common::FileEntry::from)
+    // and writeln! already adds LF.
+    for line in lines[..lines.len() - 1].iter() {
+        writeln!(file, "{}", line.raw_string)?;
+    }
+
+    Ok(())
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::env::temp_dir;
-    use std::fs::{remove_file, File};
+    use crate::common::FileEntry;
+    use std::fs::{self, remove_file, File};
 
     fn run_relative_path_asserts(assertions: Vec<(&str, &str, &str)>) {
         for (target, base, relative) in assertions {
@@ -76,7 +92,8 @@ mod tests {
     #[test]
     #[cfg(not(windows))]
     fn test_canonicalize() {
-        let path = temp_dir().join(".env");
+        let dir = tempfile::tempdir().expect("create temp dir");
+        let path = dir.path().join(".env");
         File::create(&path).expect("file created");
 
         let fs_canonical_path = std::fs::canonicalize(&path).expect("canonical path by std::fs");
@@ -94,7 +111,8 @@ mod tests {
         const UNC_PREFIX: &str = "\\\\?\\";
 
         let file_name = String::from(".env");
-        let path = temp_dir().join(&file_name);
+        let dir = tempfile::tempdir().expect("create temp dir");
+        let path = dir.path().join(&file_name);
         File::create(&path).expect("create testfile");
 
         let dunce_canonical_path = dunce::canonicalize(&path).expect("canonical path by `dunce`");
@@ -108,6 +126,45 @@ mod tests {
         // The result of `fs_utils` must match `dunce` on Windows
         assert_eq!(canonical_path, dunce_canonical_path);
         assert!(!contains_unc);
+
+        remove_file(path).expect("temp file deleted");
+    }
+
+    #[test]
+    fn write_file_test() {
+        let file_name = String::from(".env");
+        let dir = tempfile::tempdir().expect("create temp dir");
+        let path = dir.path().join(&file_name);
+
+        let fe = FileEntry {
+            path: path.clone(),
+            file_name,
+            total_lines: 3,
+        };
+
+        let lines = vec![
+            LineEntry {
+                number: 1,
+                file: fe.clone(),
+                raw_string: String::from("A=B"),
+            },
+            LineEntry {
+                number: 2,
+                file: fe.clone(),
+                raw_string: String::from("Z=Y"),
+            },
+            LineEntry {
+                number: 3,
+                file: fe,
+                raw_string: String::from("\n"),
+            },
+        ];
+
+        assert!(write_file(&path, lines).is_ok());
+        assert_eq!(
+            b"A=B\nZ=Y\n",
+            fs::read(path.as_path()).expect("file read").as_slice()
+        );
 
         remove_file(path).expect("temp file deleted");
     }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,6 +5,7 @@ use std::path::PathBuf;
 
 mod checks;
 mod common;
+mod fixes;
 mod fs_utils;
 
 pub use checks::available_check_names;
@@ -35,6 +36,7 @@ pub fn run(args: clap::ArgMatches, current_dir: &PathBuf) -> Result<Vec<Warning>
         file_paths.extend(get_file_paths(input_paths, &excluded_paths, is_recursive));
     }
 
+    let is_fix = args.is_present("fix");
     let mut warnings: Vec<Warning> = Vec::new();
 
     for path in file_paths {
@@ -43,14 +45,18 @@ pub fn run(args: clap::ArgMatches, current_dir: &PathBuf) -> Result<Vec<Warning>
             None => continue,
         };
 
-        let file_with_lines = match FileEntry::from(relative_path) {
+        let (fe, strs) = match FileEntry::from(relative_path) {
             Some(f) => f,
             None => continue,
         };
 
-        let lines = get_line_entries(file_with_lines.0, file_with_lines.1);
+        let mut lines = get_line_entries(&fe, strs);
 
-        let result = checks::run(lines, &skip_checks);
+        let mut result = checks::run(&lines, &skip_checks);
+        if is_fix && fixes::run(&mut result, &mut lines) > 0 {
+            fs_utils::write_file(&fe.path, lines)?;
+        }
+
         warnings.extend(result);
     }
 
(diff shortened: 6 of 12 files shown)


mgrachev:
@evgeniy-r Thank you for your contribution! 🚀

This PR has a lot of changes. I guess I need more time to review this feature.

mgrachev:
> I added the field with a checker name to `Warning` (it is possible to determine the checker by the text of the message, but I don't like this approach).

I'd like to minimize the changes in this PR, so I suppose to transfer this item to another PR.

DDtKey:
Thanks for the contribution, autofix are a really important part of development :rocket: 

I would like to consider this PR a little more when I will have a bit more free time.

It seems important to me to discuss the moment, with a fix using the `Warnings`.
The ability to categorize as “fixed” and “not fixed” really _looks great!_ :fire: 

But some questions worry me:
1) Will this API accurately make it easy to scale the application (add new fixes for checks)?
2) Is there enough information in the warnings to correct various types of warnings?

Perhaps the fixers should still be able to work **without** a preliminary check, what do you think?

The fact is that we are unlikely to be able to use the information, let's say about the sorting order, we’ll just have to fix the whole file. And roughly speaking the `Warning` information gives not so much in such cases. It also possibly affects performance, although not critically.

Maybe I'm completely wrong, but I want to raise this issue to look into the distance :thinking: 

@dotenv-linter/core

evgeniy-r:
> I'd like to minimize the changes in this PR, so I suppose to transfer this item to another PR.

I'm agree, I will do it in a couple of days.

evgeniy-r:
@DDtKey, thank you for feedback!

> Will this API accurately make it easy to scale the application (add new fixes for checks)?

Important question. Yes, I experimented a bit and I'm sure that it will be possible to implement all existing checks based on this approach.

> Is there enough information in the warnings to correct various types of warnings?

Sometimes (in several checks) we will work with the entire line entry collection and will use warnings only for indication (`Fixed`).
These checks (preliminarily): `DuplicatedKey`, `ExtraBlankLine`, `UnorderedKey`.

> Perhaps the fixers should still be able to work without a preliminary check, what do you think?

I’ll think it over, but it seems to me that in terms of performance, checking first and then fixing is not a problem. On the other hand, we clearly link checks and fixes now.

> The fact is that we are unlikely to be able to use the information, let's say about the sorting order, we’ll just have to fix the whole file. 

For some checks, this is true, for others - not.

DDtKey:
I apologize, I deleted the question before I received the answer, because I answered myself to some extent :disappointed: 

I will return it to leave a story:
> Thanks for the contribution, autofix are a really important part of development :rocket: 
> 
> I would like to consider this PR a little more when I will have a bit more free time.
> 
> It seems important to me to discuss the moment, with a fix using the `Warnings`.
> The ability to categorize as “fixed” and “not fixed” really _looks great!_ :fire: 
> 
> But some questions worry me:
> 1) Will this API accurately make it easy to scale the application (add new fixes for checks)?
> 2) Is there enough information in the warnings to correct various types of warnings?
> 
> Perhaps the fixers should still be able to work without a preliminary check, what do you think?
> 
> The fact is that we are unlikely to be able to use the information, let's say about the sorting order, we’ll just have to fix the whole file. And roughly speaking the `Warning` information gives not so much in such cases. It also possibly affects performance, although not critically.
> 
> At the same time, it is absolutely correct to display warnings that we cannot correct or are ambiguous (for example: `IncorrectDelimiterChecker`).
> 
> Maybe I'm completely wrong, but I want to raise this issue to look into the distance :thinking: 


@evgeniy-r Thanks for the detailed answer, I agree that the approach justifies itself.

evgeniy-r:
I have splitted this PR into two (https://github.com/dotenv-linter/dotenv-linter/pull/234).
I can't push the branch to this repo and rebase this PR to another branch, so there are 2 commits for now (one for each PR).

mgrachev:
@evgeniy-r The PR https://github.com/dotenv-linter/dotenv-linter/pull/234 has been merged. Please, update this PR.

evgeniy-r:
I have updated PR.

README.md
```diff
@@ -209,6 +209,15 @@ TrailingWhitespace
 UnorderedKey
 ```
 
+`dotenv-linter` can also automatically fix warnings in the files. Currently only one kind of warnings is fixed
+(`LowercaseKey`). You should use the argument `--autofix` (or its short version `-f`) for this.
+
+```shell script
+$ dotenv-linter -f
+Unfixed: .env:2 DuplicatedKey: The BAR key is duplicated
+  Fixed: .env:3 LowercaseKey: The FOO key should be in uppercase
```
mgrachev, line 218: To tell the truth, I don't like this output.
It will look a little ugly if we check many files:
```shell script
Unfixed: .env:1 KeyWithoutValue: The FOO key should be with a value or have an equal sign
Unfixed: .env:2 LeadingCharacter: Invalid leading character detected
Unfixed: .env:2 KeyWithoutValue: The  BAR key should be with a value or have an equal sign
Unfixed: .env:3 LeadingCharacter: Invalid leading character detected
Unfixed: .env:3 KeyWithoutValue: The 1 key should be with a value or have an equal sign
Unfixed: subdir/.env:1 LeadingCharacter: Invalid leading character detected
Unfixed: subdir/.env:1 KeyWithoutValue: The 2 key should be with a value or have an equal sign
  Fixed: subdir/.env:2 LowercaseKey: The foo key should be in uppercase
Unfixed: subdir/.env:2 QuoteCharacter: The value is wrapped in quotes
Unfixed: subdir/.env.common:2 ExtraBlankLine: Extra blank line detected
  Fixed: subdir/.env.common:3 LowercaseKey: The foo key should be in uppercase
Unfixed: subdir/.env.common:5 ExtraBlankLine: Extra blank line detected
Unfixed: subdir/test.env:2 UnorderedKey: The FOO key should go before the _TEST key
  Fixed: subdir/test.env:3 LowercaseKey: The var key should be in uppercase
```

I'd like to see something like this:
```
Fixed warnings:
subdir/.env:2 LowercaseKey: The foo key should be in uppercase
subdir/.env.common:3 LowercaseKey: The foo key should be in uppercase
subdir/test.env:3 LowercaseKey: The var key should be in uppercase

Unfixed warnings:
.env:1 KeyWithoutValue: The FOO key should be with a value or have an equal sign
.env:2 LeadingCharacter: Invalid leading character detected
.env:2 KeyWithoutValue: The  BAR key should be with a value or have an equal sign
.env:3 LeadingCharacter: Invalid leading character detected
.env:3 KeyWithoutValue: The 1 key should be with a value or have an equal sign
subdir/.env:1 LeadingCharacter: Invalid leading character detected
subdir/.env:1 KeyWithoutValue: The 2 key should be with a value or have an equal sign
subdir/.env:2 QuoteCharacter: The value is wrapped in quotes
subdir/.env.common:2 ExtraBlankLine: Extra blank line detected
subdir/.env.common:5 ExtraBlankLine: Extra blank line detected
subdir/test.env:2 UnorderedKey: The FOO key should go before the _TEST key
```

src/common.rs
```diff
@@ -4,9 +4,10 @@ use std::path::PathBuf;
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct Warning {
-    check_name: String,
+    pub check_name: String,
     line: LineEntry,
     message: String,
+    pub is_fixed: Option<bool>,
```
mgrachev, line 10:
```suggestion
    pub is_fixed: bool,
```
What do you think about using just `bool` type instead of `Option<bool>`?

src/common.rs
```diff
@@ -16,16 +17,40 @@ impl Warning {
...
             message,
+            is_fixed: None,
         }
     }
+
+    pub fn line_number(&self) -> usize {
+        self.line.number
+    }
+
+    pub fn set_fixed(&mut self, val: bool) {
```
mgrachev, line 28:
```suggestion
    pub fn mark_as_fixed(&mut self, val: bool) {
```
I like more this name 🙂

src/common.rs
```diff
@@ -16,16 +17,40 @@ impl Warning {
...
 }
 
 impl fmt::Display for Warning {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let fixed_prefix = match self.is_fixed {
+            Some(is_fixed) => {
+                if is_fixed {
+                    "  Fixed: "
+                } else {
+                    "Unfixed: "
```
mgrachev, line 40: I wouldn't like to print this prefix before all warnings.
It would be great if we grouped warnings into 2 groups - Fixed and Unfixed.

src/file_writer.rs
```diff
@@ -0,0 +1,63 @@
+use std::fs::File;
+use std::io::{self, Write};
+use std::path::PathBuf;
+
+use crate::common::LineEntry;
+
+// I think that we should create a backup copy
+pub fn write(path: &PathBuf, lines: Vec<LineEntry>) -> io::Result<()> {
```
mgrachev, line 8: We already have the `fs_utils.rs` file. Please, transfer this method there.

src/file_writer.rs
```diff
@@ -0,0 +1,63 @@
...
+    let mut file = File::create(path)?;
+
+    for line in lines[..lines.len() - 1].iter() {
```
mgrachev, line 11: Please, leave here a comment about `lines[..lines.len() - 1]`.

src/file_writer.rs
```diff
@@ -0,0 +1,63 @@
...
+                number: 3,
+                file: fe,
+                raw_string: String::from("\n"),
+            },
+        ];
+
+        assert!(write(&path, lines).is_ok());
+        assert_eq!(
+            b"A=B\nZ=Y\n",
+            fs::read(path.to_str().unwrap()).unwrap().as_slice()
```
mgrachev, line 58: We don't use the `unwrap` method.

src/fixes.rs
```diff
@@ -0,0 +1,147 @@
...
+    fn name(&self) -> &str;
+
+    fn fix_warnings(
+        &self,
+        warnings: Vec<&mut Warning>,
+        lines: &mut Vec<LineEntry>,
+    ) -> Option<usize> {
+        let mut count: usize = 0;
+        for warning in warnings {
+            let line = &mut lines[warning.line_number() - 1];
```
mgrachev, line 15: Can use the `get_mut` method here?

src/fixes.rs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        Some(count)
+    }
+
+    fn fix_line(&self, _line: &mut LineEntry) -> Option<()> {
+        None
+    }
+
+    fn set_fixed_on_all(&self, warnings: Vec<&mut Warning>) -> Option<usize> {
```
mgrachev, line 29: This method is not used?
If so, please remove it.

README.md
DDtKey: Should we print corrected `warings`? :thinking: 
Linters often do not display this information, but since they work inside the version control system, I believe that it is not always relevant for us

But I too prefer the option of dividing information into blocks mentioned by  @mgrachev :rocket:

src/main.rs
```diff
@@ -15,11 +15,21 @@ fn main() -> Result<(), Box<dyn Error>> {
     }
 
     let warnings = dotenv_linter::run(args, &current_dir)?;
-    if warnings.is_empty() {
+
+    let mut all_fixed = true;
```
DDtKey, line 19: It seems to me, for Rust, it will be more idiomatic to use functional programming approaches instead of a mutable variable (`filter()`, `map()`,  `filter_map()`, `any()` and others) :eyes:

src/fixes.rs
```diff
@@ -0,0 +1,147 @@
...
+    if warnings.is_empty() {
+        return 0;
+    }
+
+    for warning in warnings.iter_mut() {
+        warning.set_fixed(false);
+    }
+
+    let mut count = 0;
+    for fixer in fixlist() {
```
DDtKey, line 62: What do you think about using a `HashMap<String, Box<dyn Fix>>` instead of a collection, which will reduce the number of nested iterations?

README.md
mgrachev: Hmm, it's a good question, because we plan to fix all warnings.
Maybe, in the future we won't print the `Unfixed` group at all. 🤔

src/fixes.rs
```diff
@@ -0,0 +1,147 @@
...
+        // We can optimize it: create check_name:warnings map in advance
+        let fixer_warnings: Vec<&mut Warning> = warnings
+            .iter_mut()
+            .filter(|w| w.check_name == fixer.name())
+            .collect();
+        if !fixer_warnings.is_empty() || fixer.name() == "ExtraBlankLine" {
```
DDtKey, line 68: Why `ExtraBlankLine` has specific condition?
I just don’t understand a bit :confused:

README.md
```diff
@@ -209,6 +209,15 @@ TrailingWhitespace
 UnorderedKey
 ```
 
+`dotenv-linter` can also automatically fix warnings in the files. Currently only one kind of warnings is fixed
+(`LowercaseKey`). You should use the argument `--autofix` (or its short version `-f`) for this.
```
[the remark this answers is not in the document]
evgeniy-r: I'm agree - it is shorter )

src/fixes.rs
```diff
@@ -0,0 +1,147 @@
...
+    ]
+}
+
+pub fn run(warnings: &mut [Warning], lines: &mut Vec<LineEntry>) -> usize {
+    if warnings.is_empty() {
+        return 0;
+    }
+
+    for warning in warnings.iter_mut() {
+        warning.set_fixed(false);
```
DDtKey, line 58: I think it's enough to make the default field `false` instead of `Option` 
It seems like unspoken logic

README.md
evgeniy-r: I think the output in groups is not hard to implement.
But the current format is easier to handle with another command, (with pipeline).

evgeniy-r: If it is matter.

src/common.rs
evgeniy-r: `None` is used when we are not fixing Warnings, in this case they are no `Fixed`, no `Unfixed`.

src/lib.rs
```diff
@@ -43,15 +46,19 @@ pub fn run(args: clap::ArgMatches, current_dir: &PathBuf) -> Result<Vec<Warning>
...
         };
 
-        let lines = get_line_entries(file_with_lines.0, file_with_lines.1);
+        let mut lines = get_line_entries(&fe, strs);
 
-        let result = checks::run(lines, &skip_checks);
+        let result = checks::run(&lines, &skip_checks);
         warnings.extend(result);
+
+        if autofix && fixes::run(&mut warnings, &mut lines) > 0 {
```
DDtKey, line 59: It seems to me, by running corrections after the `warnings.extend(result)`, in fact we restart the check for already fixed ones (just unnecessary iterations)
Perhaps this is more correct to do on the `result`?

```suggestion
        if autofix && fixes::run(&mut result, &mut lines) > 0 {
```

src/common.rs
evgeniy-r: There are pretty many `set_` methods in Rust std library )
https://doc.rust-lang.org/std/net/enum.SocketAddr.html#method.set_ip
https://doc.rust-lang.org/std/env/fn.set_var.html
https://doc.rust-lang.org/std/fs/struct.File.html#method.set_len

But `mark_as_fixed` is good one, too.

src/common.rs
evgeniy-r: I will remove it, if we change the output format.

src/file_writer.rs
evgeniy-r: OK, and how about backup copy?
Or we can just mention this in the help.

src/file_writer.rs
evgeniy-r: It is test, but OK.

src/fixes.rs
evgeniy-r: Maybe, I'll take a look.

src/file_writer.rs
evgeniy-r: OK

src/main.rs
evgeniy-r: I am agree about the functional style.
But first we need to finally decide how we will display warnings (format).

src/fixes.rs
evgeniy-r: Yes, this is remains from some of the experiments )

src/file_writer.rs
mgrachev: We can do it  later.

src/fixes.rs
evgeniy-r: Right now, we can (and maybe should) remove it.
But some fixes (namely `UnorderedKey`) can spawn extra blank lines (depends on implementation).
I experimented a bit with implementing various fixes when designing the software interface of fixes.

evgeniy-r: I will remove it for now.

src/fixes.rs
evgeniy-r: `None` is used when we are not fixing Warnings, in this case they are no Fixed, no Unfixed.

src/lib.rs
evgeniy-r: It seems you found a bug )
I will check.

README.md
mgrachev: > I think the output in groups is not hard to implement.

Great. I'd then stay on it.

evgeniy-r:
@mgrachev, @DDtKey thank you very much for your feedback.
I will make changes.

src/fixes.rs
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    let mut count = 0;
+    for fixer in fixlist() {
+        // We can optimize it: create check_name:warnings map in advance
```
DDtKey, line 63: It would be very good to optimize this, the `group_by` function + storing the `fixlist` in the form of a `HashMap<String, Box<dyn Fix>>` would help a lot.

We can use lightweight [itertools](https://crates.io/crates/itertools) for this, because it really will allow to distribute all the `Warnings` in their checker names in 1 iteration, but [it has specifics](https://docs.rs/itertools/0.9.0/itertools/trait.Itertools.html#method.into_group_map):

```
let warning_groups = warnings
   .iter_mut()
   .map(|w| (w.name(), w))
   .into_group_map();

for (k, v) in warnings_map.inter() {
   let fixer = fixers_map.get(k);
   // ... fix warnings 
}
```

But I think someone else needs an opinion on this :thinking: 
@dotenv-linter/core, what do you think?

src/common.rs
DDtKey: > `None` is used when we are not fixing Warnings, in this case they are no `Fixed`, no `Unfixed`.

It seems to me, by default, an `Warning` cannot have such a state.
When created, it is not automatically fixed.
Correct me if it is not

src/common.rs
DDtKey: > There are pretty many `set_` methods in Rust std library )
> https://doc.rust-lang.org/std/net/enum.SocketAddr.html#method.set_ip
> https://doc.rust-lang.org/std/env/fn.set_var.html
> https://doc.rust-lang.org/std/fs/struct.File.html#method.set_len
> 
> But `mark_as_fixed` is good one, too.

I also think the `mark_as_fixed` is more clear, because it's a setter for a boolean variable :)

src/common.rs
evgeniy-r: Yes, it is no automatically Unfixed, no Fixed. This is the point.
But I will study alternative approaches.

src/fixes.rs
DDtKey: > `None` is used when we are not fixing Warnings, in this case they are no Fixed, no Unfixed.

Just anyway, we see them (`None`) as false here :thinking:

evgeniy-r: true - `Fixed` prefix (we tried and fixed)
false - `Unfixed` prefix (we tried and couldn't fix)
None - no prefix (we didn't try)

evgeniy-r: But as I say, I will take a look on alternative approaches.

README.md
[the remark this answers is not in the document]
evgeniy-r: I renamed the argument, also made some associated renames (for more consistent naming).

src/file_writer.rs
evgeniy-r: I moved it to `fs_utils`.

src/file_writer.rs
evgeniy-r: Done

src/file_writer.rs
evgeniy-r: I replaced `unwrap` (here and in one more place).

src/fixes.rs
evgeniy-r: I removed it.

src/fixes.rs
evgeniy-r: Done

src/lib.rs
evgeniy-r: I fixed this and also added a test for fixing multiple files.

src/common.rs
evgeniy-r: I have renamed this method.

src/fixes.rs
evgeniy-r: We can, but then we have to use `expect` (the line at this index must exist, if not, we should panic):
```rust
let line = lines.get_mut(warning.line_number() - 1).expect("get line");
```

The first version is shorter.

evgeniy-r: I think `get_mut` is more suitable for cases where it is not known whether an element exists.

evgeniy-r: https://doc.rust-lang.org/std/vec/struct.Vec.html#indexing
> Use get and get_mut if you want to check whether the index is in the Vec.

README.md
evgeniy-r: I have changed the output format.
Also I have added some tests for this feature.

src/common.rs
evgeniy-r: I have changed the output format.

src/main.rs
evgeniy-r: I changed the output format and there is no `all_fixed` now.

evgeniy-r:
I changed the output format and there is no need in `Option<bool>` anymore (since the output format of each message no longer depends on the fixing option). I will simplify the code.

evgeniy-r:
Also there is some problem with tests on Windows platform - I will look in more detail.

evgeniy-r:
> 
> 
> Also there is some problem with tests on Windows platform - I will look in more detail.

I have fixed tests, everything is OK.

src/fixes.rs
evgeniy-r: I changed the output format and there is no need in Option<bool> anymore.

src/common.rs
evgeniy-r: There is no need in Option anymore.

src/fixes.rs
evgeniy-r: I can do this optimization now, or we can leave this for a separate issue.
I think there is no need for additional libraries.

evgeniy-r:
I seem to have answered all the comments.
I can rebase on the master branch, if it is needed.

src/fixes.rs
mgrachev: Let's leave it for the future. Maybe, someone else will do it.
And I wouldn't want to add an additional dependency.

src/common.rs
```diff
@@ -16,8 +17,17 @@ impl Warning {
...
             message,
+            is_fixed: false,
         }
     }
+
+    pub fn line_number(&self) -> usize {
+        self.line.number
+    }
+
+    pub fn mark_as_fixed(&mut self, val: bool) {
```
DDtKey, line 28: I apologize, it seemed to me that we are only setting the meaning of `true`, so such a renaming would make sense.
That is, it seems to me that this setter should not take a value, but simply set true independently.

If need set false, `set_fixed` really more suitable. Or add second method `mark_as_unfixed` :)

evgeniy-r: There is no `Option` anymore, so it seems that we can remove the argument (because the default value is `false`).

evgeniy-r: Done.

DDtKey:
You did a great job! :+1:
Thank you!  🙏
This will be great ground for release v2.2.0, with the primary goal of implementing the remaining fixers 📄

src/fixes.rs
mgrachev: I'd insist to use the `get_mut` method to escape panics:
> However be careful: if you try to access an index which isn't in the Vec, your software will panic!

README.md
```diff
@@ -209,6 +209,19 @@ TrailingWhitespace
...
+(`LowercaseKey`). You should use the argument `--fix` (or its short version `-f`) for this.
+
+```shell script
+$ dotenv-linter -f
+Fixed warnings:
+.env:3 LowercaseKey: The foo key should be in uppercase
+
+Unfixed warnings:
+.env:2 DuplicatedKey: The BAR key is duplicated
+
```
mgrachev, line 222: Extra blank line detected 😉

src/fixes.rs
evgeniy-r: But what can we do except panic in this situation (when there is wrong line number in the warning)?
If we use `expect`, we get panic too.

DDtKey: > But what can we do except panic in this situation (when there is wrong line number in the warning)?
> If we use `expect`, we get panic too.

May be return `None` instead use `expect`? What do you think?

evgeniy-r: If the entry is not in the array, it means something unexpected happened, the program does not work correctly, we can’t just ignore it.

DDtKey: > If the entry is not in the array, it means something unexpected happened, the program does not work correctly, we can’t just ignore it.
> 
> 

Hmm :thinking:
This is a fix that failed, as I understand it.  We can display some information about this, but not complete the work with panic.

Just leaving the uncorrected lines as they are, although maybe this is a bad solution.

src/main.rs
```diff
@@ -14,12 +14,38 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-        process::exit(0);
+
+    if is_fix {
+        if warnings.iter().any(|w| w.is_fixed) {
+            println!("Fixed warnings:");
+            warnings
+                .iter()
+                .filter(|w| w.is_fixed)
+                .for_each(|w| println!("{}", w));
+            println!();
```
mgrachev, line 28: Why is this needed? 🤔

src/main.rs
```diff
@@ -14,12 +14,38 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+        }
+
+        if warnings.iter().any(|w| !w.is_fixed) {
+            println!("Unfixed warnings:");
+            warnings
+                .iter()
+                .filter(|w| !w.is_fixed)
+                .for_each(|w| println!("{}", w));
+            println!();
```
mgrachev, line 37: The same question 🤔

README.md
evgeniy-r: There is a blank line after a warning group in your example :slightly_smiling_face::

> Fixed warnings:
> subdir/.env:2 LowercaseKey: The foo key should be in uppercase
> subdir/.env.common:3 LowercaseKey: The foo key should be in uppercase
> subdir/test.env:3 LowercaseKey: The var key should be in uppercase
> 
> Unfixed warnings:
> .env:1 KeyWithoutValue: The FOO key should be with a value or have an equal sign
> .env:2 LeadingCharacter: Invalid leading character detected
> .env:2 KeyWithoutValue: The  BAR key should be with a value or have an equal sign

Should I remove it?

evgeniy-r: Or leave a blank line between groups, but remove it at the end?

src/fixes.rs
evgeniy-r: The missing index means corrupted data or an error in program logic, not failed fix.
It must always be present, if everything is correct (because warnings contain correct line numbers).

DDtKey: Yes, I understand and agree, but I will explain what I mean:
fixes is an additional feature and in the case of a program error, it would seem not bad to save the output of the checks results. 
And additionally printing an error, for example, that the fixes encountered a problem.

It might make sense to think about other options.
Just giving panic to the end user of the cli tool will not be very useful and understandable for him.

README.md
mgrachev: Please, remove extra blank line at the end.

evgeniy-r: Done

evgeniy-r:
Sometimes I see that tests on Windows failed:
https://github.com/dotenv-linter/dotenv-linter/pull/228/checks?check_run_id=862439701

(problem with deleting a temporary file)

src/fixes.rs
evgeniy-r: If you specified the `--fix` flag, you definitely want the files to be fixed, I don't think this is an additional feature in such case.

About panic, I will try to give an analogy (a little exaggerated):
```rust
let mut i = 0;
let v = vec![1, 2, 3, 4, 5];

while i < v.len() {
    // should we panic, if there is no element?
    let item = v[i];

    i += 1;
}
```

evgeniy-r:
@mgrachev, I have given different names to the temp files in `fs_utils` tests (but I'm not sure if this is causing the error).

mgrachev:
@DDtKey Can you help @evgeniy-r with tests on Windows?

DDtKey:
> @DDtKey Can you help @evgeniy-r with tests on Windows?

Yes, of course!

As soon as I am get free time, I will definitely check.

> @mgrachev, I have given different names to the temp files in `fs_utils` tests (but I'm not sure if this is causing the error).

I think this is due to the `temp_dir` method.  It works differently than `tempdir` and does not give the same lifetime guarantees.

evgeniy-r:
https://doc.rust-lang.org/std/env/fn.temp_dir.html

**Windows**
Returns the value of, in order, the TMP, TEMP, USERPROFILE environment variable if any are set and not the empty string. Otherwise, `temp_dir` returns the path of the Windows directory.

DDtKey:
> https://doc.rust-lang.org/std/env/fn.temp_dir.html
> 
> **Windows**
> Returns the value of, in order, the TMP, TEMP, USERPROFILE environment variable if any are set and not the empty string. Otherwise, `temp_dir` returns the path of the Windows directory.

Yes, but `File::create` also does not change the file name according to the specification.

Among the action logs, I noticed a `AccessDenied`.  Therefore, I just assumed.  Now, unfortunately, I can’t look at it from a PC, but I’ll definitely check later.

evgeniy-r:
I was able to get PC )
Before renaming: 7 fails from 20 test runs.
After renaming: 0 fails from 20 test runs.

DDtKey:
> I was able to get PC )
> Before renaming: 7 fails from 20 test runs.
> After renaming: 0 fails from 20 test runs.

Sorry, I mean, it's hard for me to comment, because I'm writing now from a smartphone 😔

This is interesting, because this has not happened with tests in the master branch so far, both on a personal laptop with Windows and in github action :thinking:

src/fixes.rs
DDtKey: In fact, as I said, I agree :)
Rust provides this protection for us.

I rather worried about the possibility of a human-readable error output, for users who do not know the internal structure of the `dotenv-linter`.

evgeniy-r:
There is only one such test in the master branch.

You should look at this commit:
https://github.com/dotenv-linter/dotenv-linter/pull/228/commits/893d5308539999338e004ca44467e3c009afa600

DDtKey:
> There is only one such test in the master branch.


I think this does not change much, but I know two working on a similar principle: [second example](https://github.com/dotenv-linter/dotenv-linter/blob/e947f60b1ae424c08f1e3c17a8ca1cc651bd29eb/src/common.rs#L187)

DDtKey:
> I was able to get PC )
> Before renaming: 7 fails from 20 test runs.
> After renaming: 0 fails from 100+ test runs.

Perhaps the matter is that tests are performed in parallel?
One test create a file with a specific name, another, begins to read or delete it?  We create a file in a temporary folder, but the same (`temp_dir()`).
While `tmpdir` creates a unique temporary folder.

evgeniy-r:
And there is no fails on Lunux.
Are you able to reproduce same results?

DDtKey:
> And there is no fails on Lunux (maybe Windows just does not delete the file immediately).
> Are you able to reproduce same results?

I was able to reproduce this.
If the tests (with `temp_dir` + same file name) are in the same module, we easily come across this on `Windows`,
and I have it **reproduced on Linux including**
Moreover, this does not happen with the `cargo test - --test-threads 1` command, which confirms the above.

DDtKey:
I think, we should use something like that:
```
        let file_name = String::from(".env");
        let tempdir = tempdir().expect("create tempdir");
        let path = tempdir.path().join(file_name);
``` 
This solves the problem of unique names and concurrency. 
How this works in integration tests.

evgeniy-r:
> I was able to reproduce this.
> If the tests (with temp_dir + same file name) are in the same test module, we easily come across this on Windows,
> and I have it reproduced on Linux including
> Moreover, this does not happen with the cargo test - --test-threads 1 command, which confirms the above.

:+1: It is an excellent finding!
I agree that we should use randomized file names.

evgeniy-r:
I added `tempdir` usage to tests. Also rebased to the fresh master.

src/fs_utils.rs
```diff
@@ -111,4 +129,43 @@ mod tests {
 
         remove_file(path).expect("temp file deleted");
```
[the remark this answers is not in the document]
evgeniy-r: I'm agree, I will make changes.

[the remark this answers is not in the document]
evgeniy-r: Done

tests/fixes/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+
+#[test]
+fn multiple_files() {
+    let testdir = TestDir::new();
+
+    let testfile1 = testdir.create_testfile("1.env", "AB=DEF\nD=BAR\n\nF=BAR\n");
+    let testfile2 = testdir.create_testfile("2.env", "abc=DEF\n\nF=BAR\n");
+    let testfile3 = testdir.create_testfile("3.env", "A=b \nab=DEF\n");
+
+    let expected_output = String::from(
```
DDtKey, line 58: Why not use `format!` with `testfile.shortname_as_str()` ?

src/main.rs
```diff
@@ -14,12 +14,36 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+            println!("\nUnfixed warnings:");
+            warnings
+                .iter()
+                .filter(|w| !w.is_fixed)
+                .for_each(|w| println!("{}", w));
+        } else {
+            process::exit(0);
+        }
+    } else {
+        if warnings.is_empty() {
```
DDtKey, line 40: I think it might be before the parent condition block, if warnings are empty - that's always ok.

evgeniy-r: "No warnings" is the "default" (likely) flow, so I made this change.

tests/fixes/mod.rs
evgeniy-r: I want more clarity in this case (we discussed this a little [here](https://github.com/dotenv-linter/dotenv-linter/pull/234#issuecomment-656341990)).

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### 🚀 Added
+- Add autofix feature (just one check for now - LowercaseKey) [#228](https://github.com/dotenv-linter/dotenv-linter/pull/228) ([@evgeniy-r](https://github.com/evgeniy-r))
```
mgrachev, line 9:
```suggestion
- Add autofix feature (for LowercaseKey check) [#228](https://github.com/dotenv-linter/dotenv-linter/pull/228) ([@evgeniy-r](https://github.com/evgeniy-r))
```
We plan to implement autofix for all checks.

README.md
```diff
@@ -209,6 +209,18 @@ TrailingWhitespace
 UnorderedKey
 ```
 
+`dotenv-linter` can also automatically fix warnings in the files. Currently only one kind of warnings is fixed
+(`LowercaseKey`). You should use the argument `--fix` (or its short version `-f`) for this.
```
mgrachev, line 213:
```suggestion
(`LowercaseKey`). You should use the argument `--fix` (or its short version `-f`) for this (will be available in [v2.2.0](https://github.com/dotenv-linter/dotenv-linter/issues/238):
```

src/fixes.rs
mgrachev: I agree with @DDtKey.
In that case we should print an error instead of panic. Maybe we should use the `Result` enum instead of `Option`.

What do you think?

evgeniy-r:
> In that case we should print an error instead of panic. Maybe we should use the Result enum instead of Option

OK, I will take a look.

src/fixes.rs
evgeniy-r: What do you think is the correct behavior in this case?

1. Print error and continue to the next warning.
2. Print error, mark all warnings from this file as unfixed again and continue to the next file.
3. Print error and exit (the current behavior).

DDtKey: If we are faced with an incorrect line number, the first option seems to be incorrect (there may be a lot of errors in the output) :thinking: 

**I like option 2**, we skip fixing a specific file, leaving only `Warnings` (because other files can be processed correctly).
But since we have to explicitly report the error, the output should look clear.

The current behavior also seems correct and acceptable if the second introduces a lot of difficulties and we will report in which file specifically the error occurred.

evgeniy-r: @mgrachev, what do you think?

mgrachev: Sorry for the late response, I was on holiday 🙂

I like the option 2 more.

evgeniy-r: OK, I will make changes )

evgeniy-r: I have added `get_mut` (option 2).

src/fixes.rs
```diff
@@ -52,8 +52,17 @@ pub fn run(warnings: &mut [Warning], lines: &mut Vec<LineEntry>) -> usize {
...
         if !fixer_warnings.is_empty() {
-            count += fixer.fix_warnings(fixer_warnings, lines).unwrap_or(0);
+            match fixer.fix_warnings(fixer_warnings, lines) {
+                Some(fixer_count) => count += fixer_count,
+                None => {
+                    for warning in warnings {
+                        warning.mark_as_unfixed();
+                    }
+                    return 0;
+                }
```
mgrachev, line 64:
```suggestion
                None => return count,
```
I think we should return the `count` variable here, because we already fixed previous lines and other warnings are unfixed by default.

mgrachev:
@evgeniy-r Thanks a lot! 🙏 You've done great job! 🔥

evgeniy-r:
You are welcome!

PR #230, opened 2020-07-03
Changed the README.md to contain only basic information

wesleimp:
#### ✔ Checklist:


README.md
```diff
@@ -53,347 +40,9 @@ Articles about dotenv-linter:
...
-FOO=BAR
-BAR=FOO
-
-✅ Correct 
-FOO=BAR
+## 📄 Documentation
 
-BAR=FOO
-```
+Documentation is live hosted at https://dotenv-linter.github.io
```
DDtKey, line 45: It seems to me that placing a table of contents documentation with links would be a good solution. 
The table of contents changes much less frequently, but will give an understanding of the presence of sections of interest.
What do you think?

mstruebing:
I don't think that a ToC in the readme would be very valuable.
What would be more valuable in my opinion would be the quickstart guide for example.
If I want to know something specific I know I have to go to  the docs, so when I come to the repository by accident or discovered it somehow I would be more interested in the general usage than how to find a specific part of the documentation.

mgrachev:
I'd like to think first what we should leave in README 🤔
I'll be back soon with an answer 🙂

mgrachev:
@wesleimp I'm sorry, but I've decided to do it myself a bit later 😕

PR #233, opened 2020-07-08
Remove `FileEntry::get_content_by_path`

### LINKED ISSUE #232: Replace FileEntry::get_content_by_path with fs::read_to_string
Replace the [FileEntry::get_content_by_path](https://github.com/dotenv-linter/dotenv-linter/blob/master/src/common.rs#L88) method with [fs::read_to_string](https://doc.rust-lang.org/std/fs/fn.read_to_string.html).

mstruebing:
in favor of `fs::read_to_string`

closes #232

#### ✔ Checklist:

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -1,6 +1,5 @@
 use std::fmt;
-use std::fs::File;
-use std::io::Read;
+use std::fs;
 use std::path::PathBuf;
 
 #[derive(Clone, Debug, PartialEq)]
@@ -48,9 +47,9 @@ impl FileEntry {
             None => return None,
         };
 
-        let content = match Self::get_content_by_path(&path) {
-            Some(s) => s,
-            None => return None,
+        let content = match fs::read_to_string(&path) {
+            Ok(content) => content,
+            Err(_) => return None,
         };
 
         let mut lines: Vec<String> = content.lines().map(|str| str.to_string()).collect();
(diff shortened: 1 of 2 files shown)


src/common.rs
```diff
@@ -48,9 +47,9 @@ impl FileEntry {
             None => return None,
         };
 
-        let content = match Self::get_content_by_path(&path) {
-            Some(s) => s,
-            None => return None,
+        let content = match fs::read_to_string(&path) {
+            Ok(content) => content,
+            Err(_) => return None,
         };
```
mgrachev, line 53:
```suggestion
        let content = fs::read_to_string(&path)?;
```
Can we use the question mark method instead of `match`?

mstruebing: That would mean we need to return some proper kind of `Result/Option`-type which respects errors in that function instead of `Option<(Self, Vec<String>)>` which leads to some further changes.
Alternatively we could extract this line into an own function and act accordingly on the result in the `from` function.

mgrachev:
@mstruebing Thanks a lot! 🚀

src/common.rs
DDtKey: Guys, I'm a little late, but actually one could use:
`let content = fs::read_to_string(&path).ok()?;`

ok () converts the `Result` to `Option`.

I apologize, without the @mgrachev comment, I did not pay attention to this.
But I think this is not so important, just for the future :relaxed:

PR #234, opened 2020-07-08
Add check_name to Warning

evgeniy-r:
I added the field with a checker name to `Warning` (it is possible to determine the checker by the text of the message, but I don't like this approach).

It is needed for the [Autofix](https://github.com/dotenv-linter/dotenv-linter/pull/228) feature.

#### ✔ Checklist:

- [x] Tests for the changes have been added (for bug fixes / features).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add build and test steps running on Windows [#216](https://github.com/dotenv-linter/dotenv-linter/pull/216) ([@mgrachev](https://github.com/mgrachev))
 
 ### 🔧 Changed
+- Add the field with a checker name to Warning [#234](https://github.com/dotenv-linter/dotenv-linter/pull/234) ([@evgeniy-r](https://github.com/evgeniy-r))
 - Remove `FileEntry::get_content_by_path` in favor of `fs::read_to_string` [#233](https://github.com/dotenv-linter/dotenv-linter/pull/233) ([@mstruebing](https://github.com/mstruebing))
 - Move show-checks flag to main.rs [#227](https://github.com/dotenv-linter/dotenv-linter/pull/227) ([@mgrachev](https://github.com/mgrachev))
 - Fix `total_lines` in some tests [#224](https://github.com/dotenv-linter/dotenv-linter/pull/224) ([@DDtKey](https://github.com/DDtKey))
--- a/src/common.rs
+++ b/src/common.rs
@@ -4,13 +4,19 @@ use std::path::PathBuf;
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct Warning {
+    pub check_name: String,
     line: LineEntry,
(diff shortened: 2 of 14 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### 🚀 Added
+- Add the field with a checker name to Warning [#234](https://github.com/dotenv-linter/dotenv-linter/pull/234) ([@evgeniy-r](https://github.com/evgeniy-r))
```
mgrachev, line 9: It's not a new feature. Please, transfer it to the `Changed` section.

evgeniy-r: Done

src/common.rs
```diff
@@ -4,22 +4,28 @@ use std::path::PathBuf;
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct Warning {
+    pub check_name: String,
```
DDtKey, line 7: Why is pub field? 
Perhaps this is necessary for another PR. I just don’t see this use right now

src/common.rs
```diff
@@ -4,22 +4,28 @@ use std::path::PathBuf;
...
     line: LineEntry,
     message: String,
 }
 
 impl Warning {
-    pub fn new(line: LineEntry, message: String) -> Self {
-        Self { line, message }
+    pub fn new(line: LineEntry, check_name: &str, message: String) -> Self {
+        let check_name = String::from(check_name);
```
DDtKey, line 14:
```suggestion

```

src/common.rs
```diff
@@ -4,22 +4,28 @@ use std::path::PathBuf;
...
+        Self {
+            line,
+            check_name,
```
DDtKey, line 17:
```suggestion
            check_name: String::from(check_name),
```

It's seems more readable for me, what do you think?

src/common.rs
evgeniy-r: Yes, in fact, this PR is splitted from another one.
I will remove `pub`.

src/common.rs
evgeniy-r: I think it is more idiomatic to use only the key in languages ​​that support this feature (like Rust).
But in this case it is not really matter, if it is more readable for most people - I will gladly change it.

evgeniy-r:
> I think we should try not use hardcode checker names in tests so often...

This is an interesting subject.
Honestly, I myself thought to create an issue about reducing the amount of scaffolding code in tests (e. g. when we create line entries).
But this is partly a matter of personal preferences: using simpler (more obvious) tests or more compact ones using various helpers. I prefer the second approach, but both have their advantages.

I think, it deserves a separate issue/PR.

src/common.rs
evgeniy-r: Done

mgrachev:
@evgeniy-r Thank you! ✌️

evgeniy-r:
You are welcome )

PR #239, opened 2020-07-14
Add controll comments

mgrachev:
#### ✔ Checklist:

- [ ] Tests for the changes have been added (for bug fixes / features);

### DIFF
--- a/src/comment.rs
+++ b/src/comment.rs
@@ -1,0 +1,51 @@
+const DL_PREFIX: &'static str = "dotenv-linter";
+const ON_PREFIX: &'static str = "on";
+const OFF_PREFIX: &'static str = "off";
+
+#[derive(Debug)]
+pub struct Comment<'a> {
+    disable: bool,
+    pub checks: Vec<&'a str>,
+}
+
+pub fn parse(s: &str) -> Option<Comment> {
+    let line_comment = s[1..].trim();
+
+    if !line_comment.starts_with(DL_PREFIX) {
+        return None;
+    }
+
+    let comments: Vec<&str> = line_comment.split(':').collect();
+
+    let right_part = comments.last()?;
+    let comments: Vec<&str> = right_part.split(' ').collect();
+
+    let (&flag, checks) = comments.split_first()?;
+
+    if flag != ON_PREFIX && flag != OFF_PREFIX {
+        return None;
+    }
+
+    let checks: Vec<&str> = checks
+        .iter()
+        .map(|&s| {
+            s.split(',')
+                .filter(|&s| !s.is_empty())
+                .collect::<Vec<&str>>()
+        })
+        .flatten()
+        .collect();
+
+    let mut disable: bool = false;
+    if flag == OFF_PREFIX {
+        disable = true;
+    }
+
+    Some(Comment { disable, checks })
+}
+
+impl Comment<'_> {
+    pub fn is_disabled(&self) -> bool {
+        self.disable
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -4,6 +4,7 @@ use std::error::Error;
 use std::path::PathBuf;
 
 mod checks;
+mod comment;
 mod common;
 mod fs_utils;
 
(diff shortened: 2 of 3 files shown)


src/comment.rs
```diff
@@ -0,0 +1,221 @@
...
+pub fn parse(s: &str) -> Option<Comment> {
+    // A comment without the # character
+    let line_comment = s[1..].trim();
+
+    if !line_comment.starts_with(PREFIX) {
+        return None;
+    }
+
+    // Split the comment into 2 parts - ["dotenv-linter", "off UnorderedKey, DuplicatedKey"]
+    let parts: Vec<&str> = line_comment.split(':').collect();
```
DDtKey, line 20: Is it better to use `strip_prefix` ? 
The split seems excess because we are not using left part.

```suggestion
    let right_parts: &str = line_comment.strip_prefix(PREFIX + ':')?;
```

src/comment.rs
```diff
@@ -0,0 +1,221 @@
...
+    // ["UnorderedKey,", "DuplicatedKey"] => ["UnorderedKey", "DuplicatedKey"]
+    // ["ExtraBlankLine,LowercaseKey"] => ["ExtraBlankLine", "LowercaseKey"]
+    let checks: Vec<&str> = checks
+        .iter()
+        .map(|&s| {
+            s.split(',')
+                .filter(|&s| !s.is_empty())
+                .collect::<Vec<&str>>()
+        })
+        .flatten()
```
DDtKey, line 41:
```suggestion
```

src/comment.rs
```diff
@@ -0,0 +1,221 @@
...
+    if flag != ON && flag != OFF {
+        return None;
+    }
+
+    // Normalize list of checks. For example:
+    // ["UnorderedKey,", "DuplicatedKey"] => ["UnorderedKey", "DuplicatedKey"]
+    // ["ExtraBlankLine,LowercaseKey"] => ["ExtraBlankLine", "LowercaseKey"]
+    let checks: Vec<&str> = checks
+        .iter()
+        .map(|&s| {
```
DDtKey, line 36:
```suggestion
        .flat_map(|&s| {
```

src/lib.rs
```diff
@@ -4,6 +4,7 @@ use std::error::Error;
 use std::path::PathBuf;
 
 mod checks;
+mod comment;
```
DDtKey, line 7: Maybe it should be placed as a submodule in the common?

DDtKey (commented):
Great job, thank you! :+1: 

I left a some comments, please see them.

src/comment.rs
```diff
@@ -0,0 +1,221 @@
...
+
+#[derive(Debug, PartialEq)]
+pub struct Comment<'a> {
+    disable: bool,
+    pub checks: Vec<&'a str>,
+}
+
+pub fn parse(s: &str) -> Option<Comment> {
+    // A comment without the # character
+    let line_comment = s[1..].trim();
```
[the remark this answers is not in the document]
mgrachev: Good catch! Thanks! 👍
Also, I'll add a test for that case.

src/comment.rs
mgrachev: Yes, you're right. Thanks for advice! 🙏

src/lib.rs
mgrachev: Definitely yes 🙂

src/common/comment.rs
```diff
@@ -0,0 +1,240 @@
...
+pub struct Comment<'a> {
+    disable: bool,
+    pub checks: Vec<&'a str>,
+}
+
+pub fn parse(s: &str) -> Option<Comment> {
+    // A comment without the # character
+    let line_comment = s.trim_start()[1..].trim();
+
+    if !line_comment.starts_with(PREFIX) {
```
DDtKey, line 15: This condition seems excess, because in 21 line now exists `strip_prefix` with `?`.
If i am not mistaken it will be enough :relaxed:

PR #240, opened 2020-08-03
Add colors and total problems to the output

wesleimp:
Add colors to the output and total problems for a better visual experience.
* `file:line` - italic
* `checkname` - bold red

The flag `--no-color` turns off the feature, leaving only the plain text

image

#### ✔ Checklist:


wesleimp:
> Hey thank you for your PR rocket
> In general a very cool idea, I like to see how many errors are there in total but I think what would be really cool would be to have this option on as default but have an argument to switch colouring off.
> That way it would be easier to process for non-humans i.e. extracting information, act on specific errors etc. or if you redirected the output into a file I think all colours and formatting will be in there ascii encoded so that the content would be hard to read.
> 
> Anyway, not wanting to make your change bad :) It's great! And I think this suggestion could be even handled in a separate PR.

@mstruebing I got your point and I like it. I'll wait for the other reviewers and do that after merge.

What do you think?

wesleimp:
> Hey thank you for your PR rocket
> In general a very cool idea, I like to see how many errors are there in total but I think what would be really cool would be to have this option on as default but have an argument to switch colouring off.
> That way it would be easier to process for non-humans i.e. extracting information, act on specific errors etc. or if you redirected the output into a file I think all colours and formatting will be in there ascii encoded so that the content would be hard to read.
> 
> Anyway, not wanting to make your change bad :) It's great! And I think this suggestion could be even handled in a separate PR.

@mstruebing thinking about you review, maybe doesn't make sense adding this feature without one way to turns it off. So, I've added the flag `--no-color` too.

src/main.rs
```diff
@@ -14,13 +15,46 @@ fn main() -> Result<(), Box<dyn Error>> {
...
     }
 
-    warnings.iter().for_each(|w| println!("{}", w));
-    process::exit(1);
+    print_warnings(args.is_present("no-color"), warnings);
+    process::exit(1)
+}
+
+fn print_warnings(no_color: bool, warnings: Vec<dotenv_linter::common::Warning>) {
+    if no_color {
```
DDtKey, line 28: It seems to be better to create two separate functions for _color output_ and _no-color_. 
This will make the code much easier to read.

src/main.rs
```diff
@@ -14,13 +15,46 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+        warnings.iter().for_each(|w| {
+            println!(
+                "{}:{} {}: {}",
+                w.line.file.path.display(),
+                w.line.number,
+                w.check_name,
+                w.message
+            )
+        });
+        println!("{} problems", warnings.len());
```
DDtKey, line 38: I doubt we need print warning count as least for `no-color`output :thinking: 
This may be redundant for CI systems and etc.

src/lib.rs
```diff
@@ -4,13 +4,13 @@ use std::error::Error;
 use std::path::PathBuf;
 
 mod checks;
-mod common;
+pub mod common;
```
DDtKey, line 7: I think it would be better to use `pub(crate)` in this place.

src/common.rs
```diff
@@ -24,8 +25,11 @@ impl fmt::Display for Warning {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(
             f,
-            "{}:{} {}: {}",
-            self.line.file, self.line.number, self.check_name, self.message
+            "{}:{} {} {}",
```
DDtKey: I would change that. 
Since at the moment this code has become unused and `color`/`no-color` is controlled via the argument.

Or I would have thought about how to do formatting at the `Warning` level without special output in the `main.rs`.

src/common.rs
```diff
@@ -4,9 +4,9 @@ use std::path::PathBuf;
...
 pub struct Warning {
-    check_name: String,
-    line: LineEntry,
-    message: String,
+    pub check_name: String,
+    pub line: LineEntry,
+    pub message: String,
 }
 
 impl Warning {
```
DDtKey, line 12: This part is out of control if we just use the output somewhere, which confuses me a little.

I looked a little at the possibilities of `colored` and it seems to me that it will be correct to use [their internal way of configuring the output](https://github.com/mackwic/colored#colorization-control) if it possible.

What do you think? :thinking:

DDtKey: Overall, I like the current implementation :+1: 

Now the changes mostly affect only the `main.rs` file and it works obviously.

wesleimp: Nice, I'll work on it

src/lib.rs
wesleimp: @DDtKey good catch, however this code has been removed

src/main.rs
```diff
@@ -14,13 +15,33 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+}
+
+fn print_total_problems(total_problems: usize) {
+    println!();
+    println!(
+        "{}",
+        format!("{} {} problems", "\u{2717}", total_problems)
+            .red()
+            .bold()
+    );
```
wesleimp, line 43: @DDtKey sorry, this refactor dismissed your review :grimacing:

DDtKey: > @DDtKey sorry, this refactor dismissed your review grimacing

Good catch! 
Thanks, it's okay :relaxed:

src/main.rs
```diff
@@ -14,13 +15,33 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+
     warnings.iter().for_each(|w| println!("{}", w));
-    process::exit(1);
+    if should_print_total_problems {
+        print_total_problems(warnings.len())
+    }
+    process::exit(1)
+}
+
+fn print_total_problems(total_problems: usize) {
```
mgrachev, line 36: I've pulled out these changes into another PR and added an additional flag not to display this message.
Please, review it https://github.com/dotenv-linter/dotenv-linter/pull/242 🙂

I think we should merge my PR #242 before this one.

mgrachev:
@wesleimp Please, resolve conflicts.

mgrachev:
@wesleimp Please, update your PR from the master branch.

mgrachev:
@wesleimp Gentle ping 🙂

mgrachev:
@wesleimp Do you have any progress on this?

mgrachev:
@wesleimp I'm sorry, but I'll close this inactive PR.

Issue #237, opened 2020-07-10
Proposal: Control comments


mgrachev:
Now we can skip checks for only all files (globally):
```
$ dotenv-linter --skip UnorderedKey EndingBlankLine
```

It would be great if we could skip comments for a specific file or line using comments:
```.env
# At the beginning - it disables checks for the whole file
# dotenv-linter:off UnorderedKey, EndingBlankLine

# dotenv-linter:off UnorderedKey (We can disable a check for only some lines)
FOO=BAR
BAR=FOO
# dotenv-linter:on UnorderedKey (And enable it again)

# Supported variants:
# dotenv-linter:off UnorderedKey (For one check)
# dotenv-linter:off UnorderedKey, EndingBlankLine (For several checks)
#dotenv-linter:off UnorderedKey,EndingBlankLine (Without spaces)
```

Control comments take precedence over the `--skip` argument. So, if we run `dotenv-linter` with the argument `--skip UnorderedKey` and we also have the comment in the `.env` file like this:
```.env
# dotenv-linter:on UnorderedKey
```
This check won't be disabled.

@dotenv-linter/core What do you think about this feature?

DDtKey:
I think this is useful control control method.
Once I thought about the ability to configure certain parameters for files via `yaml` configuration file, but it seemed to me in a bulky way for such an easy linter :sweat_smile: 
You suggestion seems better for checks and fixes inside file and this makes it possible to ignore certain lines :+1:

mgrachev:
> Once I thought about the ability to configure certain parameters for files via `yaml` configuration file, but it seemed to me in a bulky way for such an lightweight linter 😅

Maybe, in the future, we'll think about how to better implement this feature 🤔

PR #241, opened 2020-08-04
Add spaced-comment checker

wesleimp:
Add `SpacedCommentChecker` to check if there's a space after `#` in comment

#### ✔ Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### 🚀 Added
+- Add spaced-comment checker [#241](https://github.com/dotenv-linter/dotenv-linter/pull/241) ([@wesleimp](https://github.com/wesleimp))
 - Add installation CI test for Windows (via `install.sh`) [#235](https://github.com/dotenv-linter/dotenv-linter/pull/235) ([@DDtKey](https://github.com/DDtKey))
 
 ### 🔧 Changed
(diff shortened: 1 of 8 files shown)


mgrachev:
@wesleimp Thank you for your contribution! 👍

@dotenv-linter/core What do you think about this check? Will it be useful?

DDtKey:
> @wesleimp Thank you for your contribution! +1
> 
> @dotenv-linter/core What do you think about this check? Will it be useful?

Unfortunately, I think this check will be excess :disappointed: 
I do not deny that this is usually the way comments are formatted, but comment line it's that I want have a free style and with this check every time will need specify `--skip`. 
And that might make it difficult to understand test comments (specific case) #237 :thinking: 

But it's not bad to have this feature as an optional feature.
For example: `--enable SpacedCommentCheck`

Perhaps if we have the ability to control checks using a configuration file or something like that it will be easier to control.

I will be glad to hear any other opinions on this matter :eyes:  :)

mgrachev:
> Unfortunately, I think this check will be excess 😞
> I do not deny that this is usually the way comments are formatted, but comment line it's that I want have a free style.
> And that might make it difficult to understand control comments (specific case) #237 🤔
> 
> But it's not bad to have this feature as an optional feature.
> For example: `--enable SpacedCommentCheck`
> 
> Perhaps if we have the ability to control checks using a configuration file or something like that it will be easier to control.
> 
> I will be glad to hear any other opinions on this matter 👀 :)

I absolutely agree with your opinion.
But, I wouldn't like to add checks that will be disabled by default.

mgrachev:
@wesleimp I'm sorry, but we've decided that this check will be excess 😕

To avoid such a situation, please open an issue for discussion before opening a PR.

PR #257, opened 2020-08-11
Implement trait `Fix` for warnings issued by `QuoteCharacterChecker`.

### LINKED ISSUE #249: Add fixer: QuoteCharacterFixer
Implement `trait Fix` to fix warnings caused by `QuoteCharacterChecker`.
And add this `Fix` to the `fixlist` in [fixes.rs](https://github.com/dotenv-linter/dotenv-linter/blob/cf176993360215596c079873725f4a4b7a48f5b0/src/fixes.rs#L32)

```env
# .env

# Wrong
FOO="BAR"

# Correct (After running `QuoteCharacterFixer` it should look like this)
FOO=BAR
```

Implementation example: [LowercaseKeyFixer](https://github.com/dotenv-linter/dotenv-linter/blob/master/src/fixes/lowercase_key.rs)

lensvol:
Closes #249.

#### ✔ Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### 🚀 Added
 - Add fixer: KeyWithoutValueFixer [#254](https://github.com/dotenv-linter/dotenv-linter/pull/254) ([@unexge](https://github.com/unexge))
+- Add fixer: QuoteCharacterFixer [#249](https://github.com/dotenv-linter/dotenv-linter/pull/249) ([@lensvol](https://github.com/lensvol))
 - Add fixer: SpaceCharacterFixer [#253](https://github.com/dotenv-linter/dotenv-linter/pull/253) ([@DDtKey](https://github.com/DDtKey))
 - Add total problems to output and `--quiet` argument [#242](https://github.com/dotenv-linter/dotenv-linter/pull/242) ([@wesleimp](https://github.com/wesleimp), [@mgrachev](https://github.com/mgrachev))
 - Add autofix feature (for LowercaseKey check) [#228](https://github.com/dotenv-linter/dotenv-linter/pull/228) ([@evgeniy-r](https://github.com/evgeniy-r))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### 🚀 Added
 - Add autofix for TrailingWhitespace [#255](https://github.com/dotenv-linter/dotenv-linter/pull/255) ([@gregcline](https://github.com/gregcline))
 - Add fixer: KeyWithoutValueFixer [#254](https://github.com/dotenv-linter/dotenv-linter/pull/254) ([@unexge](https://github.com/unexge))
-- Add fixer: QuoteCharacterFixer [#249](https://github.com/dotenv-linter/dotenv-linter/pull/249) ([@lensvol](https://github.com/lensvol))
+- Add fixer: QuoteCharacterFixer [#257](https://github.com/dotenv-linter/dotenv-linter/pull/257) ([@lensvol](https://github.com/lensvol))
 - Add fixer: SpaceCharacterFixer [#253](https://github.com/dotenv-linter/dotenv-linter/pull/253) ([@DDtKey](https://github.com/DDtKey))
 - Add total problems to output and `--quiet` argument [#242](https://github.com/dotenv-linter/dotenv-linter/pull/242) ([@wesleimp](https://github.com/wesleimp), [@mgrachev](https://github.com/mgrachev))
 - Add autofix feature (for LowercaseKey check) [#228](https://github.com/dotenv-linter/dotenv-linter/pull/228) ([@evgeniy-r](https://github.com/evgeniy-r))
--- a/README.md
+++ b/README.md
@@ -372,7 +372,7 @@ FOO_BAR=FOOBAR
 
 ### Quote character
 
-Detects if a value is wrapped in quotes:
+Detects if a value contains quote characters (`'` / `"`):
 
 ```env
 ❌ Wrong
@@ -380,6 +380,9 @@ FOO="BAR"
 
 ❌ Wrong
 FOO='BAR'
+
+❌ Wrong
+FOO='B"AR'
 
 ✅ Correct
 FOO=BAR
--- a/src/checks/quote_character.rs
+++ b/src/checks/quote_character.rs
@@ -89,6 +89,30 @@ mod tests {
                     String::from("The value has quote characters (\', \") in it"),
                 )),
             ),
+            (
+                LineEntry {
+                    number: 2,
+                    file: FileEntry {
+                        path: PathBuf::from(".env"),
+                        file_name: ".env".to_string(),
+                        total_lines: 2,
+                    },
+                    raw_string: String::from("FOO='B\"AR'"),
+                },
+                Some(Warning::new(
+                    LineEntry {
+                        number: 2,
+                        file: FileEntry {
+                            path: PathBuf::from(".env"),
+                            file_name: ".env".to_string(),
+                            total_lines: 2,
+                        },
+                        raw_string: String::from("FOO='B\"AR'"),
+                    },
+                    "QuoteCharacter",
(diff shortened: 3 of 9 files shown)


src/fixes/quote_character.rs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+impl Fix for QuoteCharacterFixer<'_> {
+    fn name(&self) -> &str {
+        self.name
+    }
+
+    fn fix_line(&self, line: &mut LineEntry) -> Option<()> {
+        let value = line.get_value()?;
+        let stripped = value.strip_prefix("\"").unwrap_or(&*value);
```
DDtKey, line 23: In addition to the double-quotes (`"`), you should remove the single-quotes (`'`).

src/fixes/quote_character.rs
```diff
@@ -0,0 +1,94 @@
...
+        let stripped = stripped.strip_suffix("\"").unwrap_or(&*value);
```
DDtKey, line 24: What do you think about use `.replace()` instead of `.strip_suffix() / .strip_prefix()`
`QuoteCharacterChecker` creates warning when value contains quotes. 
I think it will be correct to delete them completely.
```suggestion
        let pure_val = val.replace("\"","")
            .replace("'","");
```

src/fixes/quote_character.rs
```diff
@@ -0,0 +1,94 @@
...
+        line.raw_string = format!("{}={}", line.get_key()?, stripped);
```
DDtKey, line 25:
```suggestion
        line.raw_string = format!("{}={}", line.get_key()?, pure_val);
```

src/fixes/quote_character.rs
DDtKey, line 23:
```suggestion
```

lensvol: Thank you! Missed that, will fix it.

src/fixes/quote_character.rs
lensvol: That did not seem prudent, since the original issue is about quotes _surrounding_ value and `replace` will remove them regardless of their positioning. My opinion is that sudden removal of quotes inside the value will violate the principle of least astonishment and may actually be detrimental if user expects them to be there.

What do you think?

DDtKey: It just seems to me that the `Fixer` should match the `Check`.

For example, suppose this variant of the value will also be processed incorrectly if you leave them inside:
```env
VARIABLE=text"incorrect"
```
After all, we will get
```env
VARIABLE=text"incorrect
```

Which also does not seem logical :thinking:

For similar cases we have ability to skip checks and we are already working on control comments (#237)

DDtKey: @dotenv-linter/core what do you say about this?
Should we check / fix only the wrapped values or all quotes (including inside)?

lensvol: This can be fixed by only removing suffix which is a reversed version of the prefix. But I see your point :) I will wait on the final decision by the team.

mgrachev: > It just seems to me that the `Fixer` should match the `Check`.

I completely agree with that.

> Should we check / fix only the wrapped values or all quotes (including inside)?

I think we should remove all quotes (including inside). This fixer should fix all warnings from `QuoteCharacterChecker`.

lensvol: @mgrachev Good, then should I also fix the wording on the `QuoteCharacterChecker`? Currently, the warning is stated as `"The value is wrapped in quotes"`, which is inconsistent with the code of the check itself.

mgrachev: Yes, please 👍

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### 🚀 Added
 - Add autofix for TrailingWhitespace [#255](https://github.com/dotenv-linter/dotenv-linter/pull/255) ([@gregcline](https://github.com/gregcline))
 - Add fixer: KeyWithoutValueFixer [#254](https://github.com/dotenv-linter/dotenv-linter/pull/254) ([@unexge](https://github.com/unexge))
+- Add fixer: QuoteCharacterFixer [#249](https://github.com/dotenv-linter/dotenv-linter/pull/249) ([@lensvol](https://github.com/lensvol))
```
mgrachev, line 11:
```suggestion
- Add fixer: QuoteCharacterFixer [#257](https://github.com/dotenv-linter/dotenv-linter/pull/257) ([@lensvol](https://github.com/lensvol))
```

README.md
```diff
@@ -372,7 +372,7 @@ FOO_BAR=FOOBAR
 
 ### Quote character
 
-Detects if a value is wrapped in quotes:
+Detects if a value contains quote characters (`'` / `"`):
```
mgrachev, line 375: Please also update the documentation [here](https://github.com/dotenv-linter/dotenv-linter/blob/master/docs/checks/quote_character.md).

src/checks/quote_character.rs
```diff
@@ -86,7 +86,7 @@ mod tests {
                         raw_string: String::from("FOO='BAR'"),
                     },
                     "QuoteCharacter",
-                    String::from("The value is wrapped in quotes"),
+                    String::from("The value has quote characters (\', \") in it"),
```
mgrachev, line 89: It would be nice if you wrote a test for the case with `FOO=B"AR`.

src/fixes/quote_character.rs
```diff
@@ -0,0 +1,94 @@
...
+    fn fix_line_test() {
+        let fixer = QuoteCharacterFixer::default();
+        let mut line = LineEntry {
+            number: 1,
+            file: FileEntry {
+                path: PathBuf::from(".env"),
+                file_name: ".env".to_string(),
+                total_lines: 1,
+            },
+            raw_string: String::from("foo=\'\"b\'\"ar\"\'"),
```
mgrachev, line 46:
```suggestion
            raw_string: String::from("FOO=\'\"b\'\"ar\"\'"),
```
The key should be in uppercase.

src/fixes/quote_character.rs
```diff
@@ -0,0 +1,94 @@
...
+        };
+        assert_eq!(Some(()), fixer.fix_line(&mut line));
+        assert_eq!("foo=bar", line.raw_string);
```
mgrachev, line 49:
```suggestion
        assert_eq!("FOO=bar", line.raw_string);
```
The key should be in uppercase.

src/fixes/quote_character.rs
```diff
@@ -0,0 +1,94 @@
...
+        let fixer = QuoteCharacterFixer::default();
+        let mut lines = vec![
+            LineEntry {
+                number: 1,
+                file: FileEntry {
+                    path: PathBuf::from(".env"),
+                    file_name: ".env".to_string(),
+                    total_lines: 3,
+                },
+                raw_string: String::from("foo=\"bar\'\""),
```
mgrachev, line 63:
```suggestion
                raw_string: String::from("FOO=\"bar\'\""),
```
The key should be in uppercase.

src/fixes/quote_character.rs
```diff
@@ -0,0 +1,94 @@
...
+            },
+        ];
+        let mut warning = Warning::new(
+            lines[0].clone(),
+            "QuoteCharacter",
+            String::from("The value has quote characters (\', \") in it"),
+        );
+
+        assert_eq!(Some(1), fixer.fix_warnings(vec![&mut warning], &mut lines));
+        assert_eq!("foo=bar", lines[0].raw_string);
```
mgrachev, line 91:
```suggestion
        assert_eq!("FOO=bar", lines[0].raw_string);
```
The key should be in uppercase.

src/checks/quote_character.rs
```diff
@@ -16,7 +16,7 @@ impl Default for QuoteCharacterChecker<'_> {
     fn default() -> Self {
         Self {
             name: "QuoteCharacter",
-            template: "The value is wrapped in quotes",
+            template: "The value has quote characters (\', \") in it",
```
DDtKey, line 19: For me, `in it` seems superfluous.
```suggestion
            template: "The value has quote characters (\', \")",
```

mgrachev:
@lensvol Thank you for your help! 🔥

mgrachev:
🙏 If it’s not difficult for you, please support the project - click on the star on GitHub ⭐️

lensvol:
My pleasure 🍷

Issue #246, opened 2020-08-10
Add fixer: IncorrectDelimiterFixer


DDtKey:
Implement `trait Fix` to fix warnings caused by `IncorrectDelimiterChecker`.
And add this `Fix` to the `fixlist` in [fixes.rs](https://github.com/dotenv-linter/dotenv-linter/blob/cf176993360215596c079873725f4a4b7a48f5b0/src/fixes.rs#L32)

Implementation example: [LowercaseKeyFixer](https://github.com/dotenv-linter/dotenv-linter/blob/master/src/fixes/lowercase_key.rs)

It is worth discussing the way of implementation.
For example, we can change invalid characters to `_`.

mgrachev:
> It is worth discussing the way of implementation.
> For example, we can change invalid characters to `_`.

I like this approach 👍

gillespiecd:
I'd like to take this on, based on the comment above, I can change any invalid characters to "_".

PR #254, opened 2020-08-10
Add key without value fixer

unexge:
#### ✔ Checklist:

- [x] This PR has been added to [CHANGELOG.md](https://github.com/dotenv-linter/dotenv-linter/blob/master/CHANGELOG.md) (at the top of the list);
- [x] Tests for the changes have been added (for bug fixes / features);
- [x] Docs have been added / updated (for bug fixes / features).

fixes https://github.com/dotenv-linter/dotenv-linter/issues/247

PR #258, opened 2020-08-13
IncorrectDelimiter fixer (#246)

gillespiecd:
#### ✔ Checklist:

- [ ] Tests for the changes have been added (for bug fixes / features);


src/fixes.rs
```diff
@@ -40,6 +41,7 @@ fn fixlist() -> Vec<Box<dyn Fix>> {
         Box::new(lowercase_key::LowercaseKeyFixer::default()),
         Box::new(space_character::SpaceCharacterFixer::default()),
         Box::new(trailing_whitespace::TrailingWhitespaceFixer::default()),
+        Box::new(incorrect_delimiter::IncorrectDelimiterFixer::default()),
```
DDtKey, line 44: Please arrange it according to the alphabet (we do this with checks) :)

src/fixes/incorrect_delimiter.rs
```diff
@@ -0,0 +1,100 @@
...
+}
+
+impl Fix for IncorrectDelimiterFixer<'_> {
+    fn name(&self) -> &str {
+        self.name
+    }
+
+    fn fix_line(&self, line: &mut LineEntry) -> Option<()> {
+        let key: String = line.get_key()?;
+        let fixed_key: String = key
```
DDtKey, line 23: Now I thought that we should not replace characters if the string begins or ends with them (thanks #254).
We can only call a character within keys a delimiter.
The rest of the cases are handled by other checks.

mgrachev: I'm sorry, but I don't fully understand what you mean.
Could you give an example?

DDtKey: > I'm sorry, but I don't fully understand what you mean.
> Could you give an example?

For example: 
`-BAR=FOO`
I think should be fixed by LeadingCharacter
and `FOO =BAR` should be fixed by `SpaceCharacter`
It's not a delimiter, because in second case, we will get: `FOO_=BAR` instead of `FOO=BAR`.

This can be influenced by the order, but given that checks can be skipped, I think this is not correct.

DDtKey: My mistake about `FOO =BAR` case, sorry. 
I did not notice `trim`.

Most likely I'm wrong, but for some reason this question does not let me go to the end

mgrachev: How would you implement this fixer?

DDtKey: I think it's dependent from [another PR](https://github.com/dotenv-linter/dotenv-linter/pull/259#discussion_r470599906).
Solution must match check logic :thinking:

gillespiecd: Sure, we can wait until the other PR is merged and then have the fixer match the new check logic.

baile320: FYI - see #267 for the PR updating the logic

gillespiecd: Sounds good, I will wait until that PR is merged before rebasing and updating.

DDtKey: @gillespiecd #267 was merged. Please, update your PR from master :relaxed:

gillespiecd: Updated, thanks for the reminder 🚀

src/fixes/incorrect_delimiter.rs
```diff
@@ -0,0 +1,102 @@
...
+
+impl Fix for IncorrectDelimiterFixer<'_> {
+    fn name(&self) -> &str {
+        self.name
+    }
+
+    fn fix_line(&self, line: &mut LineEntry) -> Option<()> {
+        let cleaned_key = remove_invalid_leading_chars(&line.get_key()?);
+
+        let fixed_key: String = cleaned_key
```
DDtKey, line 24: In this case `IncorrectDelimiterFixer` will fix also `LeadingCharacter` warnings but it's task of `LeadingCharacterFixer`.
What if the user specifies to ignore the `LeadingCharacter` check, but gets the fixes, :thinking: 

We should leave leading chars as it is in my opinion.  
Maybe add a test for such cases?

gillespiecd: That's a good point, I can skip those chars and then apply the delimiter fixer afterwards.

gillespiecd: I think I've come up with a solution, please take a look 🙏 . I added a test case to ensure correct behavior.

It won't handle invalid suffixes, but that should be left to another checker/fixer?

Issue #248, opened 2020-08-10
Add fixer: LeadingCharacterFixer


DDtKey:
Implement `trait Fix` to fix warnings caused by `LeadingCharacterChecker`.
And add this `Fix` to the `fixlist` in [fixes.rs](https://github.com/dotenv-linter/dotenv-linter/blob/cf176993360215596c079873725f4a4b7a48f5b0/src/fixes.rs#L32)

```env
# .env

# Wrong
*RAILS_ENV=development

# Correct (After running `LeadingCharacterFixer` it should look like this)
RAILS_ENV=development
```

Implementation example: [LowercaseKeyFixer](https://github.com/dotenv-linter/dotenv-linter/blob/master/src/fixes/lowercase_key.rs)

norraell:
I would like to take this one.

Issue #252, opened 2020-08-10
Add fixer: UnorderedKeyFixer


DDtKey:
Implement `trait Fix` to fix warnings caused by `UnorderedKeyChecker`.
And add this `Fix` to the `fixlist` in [fixes.rs](https://github.com/dotenv-linter/dotenv-linter/blob/cf176993360215596c079873725f4a4b7a48f5b0/src/fixes.rs#L32)

```env
# .env

# Wrong
RAILS_ENV=development
DEBUG=true

BAR=FOO

# Correct (After running `UnorderedKeyFixer` it should look like this)
DEBUG=true
RAILS_ENV=development

BAR=FOO
```
**Important**: _sorting within groups_

Implementation example: [LowercaseKeyFixer](https://github.com/dotenv-linter/dotenv-linter/blob/master/src/fixes/lowercase_key.rs)

evgeniy-r:
I take this one.

PR #259, opened 2020-08-14
Implement LeadingCharacterFixer

baile320:
This will close: https://github.com/dotenv-linter/dotenv-linter/issues/248

I have a couple comments/questions:
- When implementing this, I ran into issues with IncorrectDelimiter: it thought, for example, "*FOO=BAR" had an incorrect delimiter (this is ultimately fixed by LeadingCharacterFixer, as expected). IMO, that is not a delimiter at all, since delimiters should occur /between/ other characters, so I added a check for that in IncorrectDelimiterChecker (here: https://github.com/dotenv-linter/dotenv-linter/compare/master...baile320:leading-char-fixer?expand=1#diff-9d20ccc745eeab0a2cfbe87e10ee860cR27-R33), and removed an erroneous test.
- Depending on the leading characters, it is still possible to get erroneous UnorderedKeys warnings. I'm not sure the best way to handle this (I made no updates to that), and would like thoughts/suggestions on how to fix (if it should be fixed here) or if it's worthy of a separate PR. There's an illustrative example in my integration test (https://github.com/dotenv-linter/dotenv-linter/compare/master...baile320:leading-char-fixer?expand=1#diff-9aff8b00dfa64d0bf60b17a8b7fe7c20R4-R22)... as you can see, it throws no UnorderedKeys warnings, although it likely should, since after cleaning the keys are technically out of order (the issue is that the checking order includes the */./1 characters at the beginning of the keys, not the "cleaned" versions of the keys).

#### ✔ Checklist:


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### 🚀 Added
+- Add fixer: LeadingCharacter [#255](https://github.com/dotenv-linter/dotenv-linter/pull/259) ([@baile320](https://github.com/baile320))
```
mgrachev, line 9:
```suggestion
- Add fixer: LeadingCharacter [#259](https://github.com/dotenv-linter/dotenv-linter/pull/259) ([@baile320](https://github.com/baile320))
```

src/checks/incorrect_delimiter.rs
```diff
@@ -126,21 +132,6 @@ mod tests {
...
-    fn leading_space_run() {
-        let mut checker = IncorrectDelimiterChecker::default();
-        let line = LineEntry {
-            number: 1,
-            file: FileEntry {
-                path: PathBuf::from(".env"),
-                file_name: ".env".to_string(),
-                total_lines: 1,
-            },
-            raw_string: String::from(" FOO=FOOBAR"),
```
mgrachev, line 139: It would be nice if you wrote a test for the case with `*FOO=BAR`.

src/fixes/leading_character.rs
```diff
@@ -0,0 +1,252 @@
...
+    }
+}
+
+impl Fix for LeadingCharacterFixer<'_> {
+    fn name(&self) -> &str {
+        self.name
+    }
+
+    fn fix_line(&self, line: &mut LineEntry) -> Option<()> {
+        line.raw_string = line.raw_string.to_string();
```
mgrachev, line 22: I don't fully understand why you do it? Can you explain me?

mgrachev:
> * When implementing this, I ran into issues with IncorrectDelimiter: it thought, for example, "*FOO=BAR" had an incorrect delimiter (this is ultimately fixed by LeadingCharacterFixer, as expected). IMO, that is not a delimiter at all, since delimiters should occur /between/ other characters, so I added a check for that in IncorrectDelimiterChecker (here: https://github.com/dotenv-linter/dotenv-linter/compare/master...baile320:leading-char-fixer?expand=1#diff-9d20ccc745eeab0a2cfbe87e10ee860cR27-R33), and removed an erroneous test.

Good catch! Thank you! 👍
But I'd like you to write a test for that case for the `IncorrectDelimiter`.

> * Depending on the leading characters, it is still possible to get erroneous UnorderedKeys warnings. I'm not sure the best way to handle this (I made no updates to that), and would like thoughts/suggestions on how to fix (if it should be fixed here) or if it's worthy of a separate PR. There's an illustrative example in my integration test (https://github.com/dotenv-linter/dotenv-linter/compare/master...baile320:leading-char-fixer?expand=1#diff-9aff8b00dfa64d0bf60b17a8b7fe7c20R4-R22)... as you can see, it throws no UnorderedKeys warnings, although it likely should, since after cleaning the keys are technically out of order (the issue is that the checking order includes the */./1 characters at the beginning of the keys, not the "cleaned" versions of the keys).

@evgeniy-r is already working to do it (https://github.com/dotenv-linter/dotenv-linter/issues/252)

DDtKey:
> > * Depending on the leading characters, it is still possible to get erroneous UnorderedKeys warnings. I'm not sure the best way to handle this (I made no updates to that), and would like thoughts/suggestions on how to fix (if it should be fixed here) or if it's worthy of a separate PR. There's an illustrative example in my integration test (https://github.com/dotenv-linter/dotenv-linter/compare/master...baile320:leading-char-fixer?expand=1#diff-9aff8b00dfa64d0bf60b17a8b7fe7c20R4-R22)... as you can see, it throws no UnorderedKeys warnings, although it likely should, since after cleaning the keys are technically out of order (the issue is that the checking order includes the */./1 characters at the beginning of the keys, not the "cleaned" versions of the keys).
> 
> @evgeniy-r is already working on that feature (#252) muscle

And I think it is permissible to use the initially correct ordering in this integration test, since the target check does not order.

CHANGELOG.md
baile320: Sorry, typo! Fixed in this commit: https://github.com/dotenv-linter/dotenv-linter/pull/259/commits/b2b50c76a30ff494b62119aed48bacf159236c04

src/checks/incorrect_delimiter.rs
baile320: No problem, I fixed here: https://github.com/dotenv-linter/dotenv-linter/pull/259/commits/49d2a79a5c07495eaf14372f57dc844889e958f9

src/fixes/leading_character.rs
baile320: That line was unnecessary, good catch, thank you :), I removed here: https://github.com/dotenv-linter/dotenv-linter/pull/259/commits/f683b815ba876f388ce3c1262d7a16b78a3aefc3 and I also extracted this function to remove the first character of the string, just in case it was unneccesarily confusing: https://github.com/dotenv-linter/dotenv-linter/pull/259/commits/3c16dc06639e0b2bb7c68c1ddd701c4dabde3bc0

src/checks/incorrect_delimiter.rs
```diff
@@ -24,7 +24,13 @@ impl Default for IncorrectDelimiterChecker<'_> {
 impl Check for IncorrectDelimiterChecker<'_> {
     fn run(&mut self, line: &LineEntry) -> Option<Warning> {
         let key = line.get_key()?;
-        if key.trim().chars().any(|c| !c.is_alphanumeric() && c != '_') {
+
+        // delimiters occur /between/ characters, not as the initial character, so we should only check those cases
+        if key
+            .trim()
+            .starts_with(|c: char| c.is_alphabetic() || c == '_')
+            && key.trim().chars().any(|c| !c.is_alphanumeric() && c != '_')
```
DDtKey, line 32: What do you think about that case?
```
*MY-VARIABLE=test
```
it's has incorrect delimiter, but starts with `*` therefore warnings don't will be create.

baile320: Thanks for catching that. I have a fix in mind: I think we should remove **all** bad characters from the beginning of the key before we perform the IncorrectDelimiter check, does that sound reasonable?

Similarly, for the LeadingCharacterFixer, the code I have written will not turn ```***KEY=VALUE``` into ```KEY=VALUE```, it will only remove one leading character and turn it into ```**KEY=VALUE```, so I think I need to fix that that before this should be merged. I will work on that a little later today, if all of that sounds reasonable.

baile320: I added a helper function to ```common.rs``` called ```remove_all_invalid_leading_chars``` to do what I described above, and used it both in IncorrectDelimiterChecker and LeadingCharacterFixer. I added some additional tests that should have caught these as well.

DDtKey: Sorry, I have been busy last couple days.
I'll look the changes soon. 

@dotenv-linter/core what do you think? This may affect #246

src/checks/incorrect_delimiter.rs
```diff
@@ -24,8 +24,21 @@ impl Default for IncorrectDelimiterChecker<'_> {
 impl Check for IncorrectDelimiterChecker<'_> {
     fn run(&mut self, line: &LineEntry) -> Option<Warning> {
         let key = line.get_key()?;
-        if key.trim().chars().any(|c| !c.is_alphanumeric() && c != '_') {
-            return Some(Warning::new(line.clone(), self.name(), self.message(&key)));
+
+        // delimiters occur /between/ characters, not as the initial character, so we should
+        // remove all invalid leading characters before checking for incorrect delimiters
+        let cleaned_key = remove_all_invalid_leading_chars(&key);
```
mgrachev, line 30: @baile320 Please move these changes to another PR.
We should merge these changes before merging the implemenation of `LeadingCharacterFixer`.

src/checks/incorrect_delimiter.rs
mgrachev: I think that these changes in the `IncorrectDelimiterChecker` should be in another PR.

src/checks/incorrect_delimiter.rs
baile320: I moved this logic to a separate PR: #267 . I can update this PR when #267 is merged to master, since it looks like the module file structured changed a little bit.

baile320:
I merged in the changes from #267 and made some other relatively minor changes related to test names, checks, formatting. I think this PR is ready to go or be re-reviewed! Thanks!

src/fixes/leading_character.rs
```diff
@@ -0,0 +1,266 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use std::path::PathBuf;
+
+    #[test]
+    fn fix_leading_period() {
```
mgrachev, line 38: What does `period` mean in that case?

src/fixes/leading_character.rs
```diff
@@ -0,0 +1,266 @@
...
+                file_name: ".env".to_string(),
+                total_lines: 7,
+            },
+            raw_string: String::from("-1&*FOO=BAR"),
+        };
+
+        assert_eq!(Some(()), fixer.fix_line(&mut leading_number));
+        assert_eq!("FOO=BAR", leading_number.raw_string);
+    }
+    #[test]
```
mgrachev, line 126:
```suggestion
    }

    #[test]
```

src/fixes/leading_character.rs
baile320: it just means to fix an item with a leading "." ... I have other tests for some other specific characters like a leading space, asterisk/*, etc. I can remove this test or rename if it helps to make it clearer?

mgrachev: Maybe this test should be named `fix_leading_dot`?

baile320: no problem, just renamed it to that! thank you

mgrachev:

PR #260, opened 2020-08-14
add ExtraBlankLineFixer (#260)

diggymo:
#245 

Below are the points I would like you to check in particular.
- Unnecessary function (`test_command_fix_fail`) came out, was it better to delete it? 

#### ✔ Checklist:

- [x] This PR has been added to [CHANGELOG.md](https://github.com/dotenv-linter/dotenv-linter/blob/master/CHANGELOG.md) (at the top of the list);
- [x] Tests for the changes have been added (for bug fixes / features);

PR #261, opened 2020-08-14
UnorderedKey fix

evgeniy-r:
PR for https://github.com/dotenv-linter/dotenv-linter/issues/252.

#### ✔ Checklist:

### DIFF
--- a/src/fixes/unordered_key.rs
+++ b/src/fixes/unordered_key.rs
@@ -62,8 +62,7 @@ impl Fix for UnorderedKeyFixer<'_> {
                     end.replace(i + 1);
                 }
 
-                let is_last = i == lines.len() - 1;
-                if line.is_empty() || is_last || is_control_comment {
+                if line.is_empty() || line.is_last_line() || is_control_comment {
                     if let Some(end_index) = end {
                         Self::sort_part(&mut lines[start_index..end_index]);
                         end = None;
@@ -102,7 +101,7 @@ impl UnorderedKeyFixer<'_> {
         slices.sort_by_cached_key(|slice| {
             // I think, that we should modify get_key() so it will return Option<&str> instead of
             // Option<String>.
-            slice.last().expect("significant line").get_key()
+            slice.last()?.get_key()
         });
 
         let mut sorted_lines = Vec::with_capacity(part.len());
@@ -119,18 +118,20 @@ mod tests {
     use super::*;
     use crate::common::tests::line_entry;
 
-    fn get_lines(strs: Vec<&str>) -> Vec<LineEntry> {
-        let total = strs.len();
+    fn get_lines(lines: Vec<&str>) -> Vec<LineEntry> {
+        let total = lines.len();
 
-        strs.iter()
+        lines
+            .iter()
             .enumerate()
-            .map(|(i, &str)| line_entry(i + 1, total, str))
+            .map(|(i, &line)| line_entry(i + 1, total, line))
             .collect()
     }
 
-    fn get_warnings(lines: &[LineEntry], wrns: Vec<(usize, &str)>) -> Vec<Warning> {
-        wrns.into_iter()
-            .map(|(i, str)| Warning::new(lines[i].clone(), "UnorderedKey", String::from(str)))
+    fn get_warnings(lines: &[LineEntry], warnings: Vec<(usize, &str)>) -> Vec<Warning> {
+        warnings
+            .into_iter()
+            .map(|(i, line)| Warning::new(lines[i].clone(), "UnorderedKey", String::from(line)))
             .collect()
     }
 
@@ -141,9 +142,9 @@ mod tests {
         fixer.fix_warnings(warning_refs, lines)
     }
 
-    fn assert_lines(lines: &[LineEntry], strs: Vec<&str>) {
-        for (i, &str) in strs.iter().enumerate() {
-            assert_eq!(str, lines[i].raw_string.as_str());
+    fn assert_lines(result: &[LineEntry], lines: Vec<&str>) {
+        for (i, &line) in lines.iter().enumerate() {
+            assert_eq!(line, result[i].raw_string.as_str());
         }
     }
 
--- a/src/fixes/unordered_key.rs
+++ b/src/fixes/unordered_key.rs
@@ -1,0 +1,724 @@
+use super::Fix;
+use crate::common::*;
+
+pub(crate) struct UnorderedKeyFixer<'a> {
+    name: &'a str,
+}
+
+impl Default for UnorderedKeyFixer<'_> {
+    fn default() -> Self {
+        Self {
+            name: "UnorderedKey",
+        }
+    }
+}
+
+// When we sort the keys, we handle a significant line (with key) with all previous comments
+// as a whole. E. g.:
+// ```
+// B=C
+// # Comment
+// A=B
+// ```
+// will be fixed to:
+// ```
+// # Comment
+// A=B
+// B=C
+// ```
+// We check the order separately in each group of entries (groups are separated by blank lines or
+// control comments).
+impl Fix for UnorderedKeyFixer<'_> {
+    fn name(&self) -> &str {
+        self.name
+    }
+
+    fn fix_warnings(
+        &mut self,
+        warnings: Vec<&mut Warning>,
+        lines: &mut Vec<LineEntry>,
+    ) -> Option<usize> {
+        // We find all sorting groups and sort them
+
+        let mut start_index = 0;
+        let mut end = None;
+        let mut is_disabled = false;
+
+        for i in 0..lines.len() {
+            let line = &lines[i];
+
+            let mut is_control_comment = false;
+            let mut controls_this_check = false;
+            let mut is_off = false;
+
+            if let Some(comment) = line.get_control_comment() {
+                is_control_comment = true;
+                controls_this_check = comment.checks.contains(&self.name);
+                is_off = comment.is_disabled();
+            }
+
+            if !is_disabled {
+                if !line.is_empty_or_comment() {
+                    end.replace(i + 1);
+                }
+
+                let is_last = i == lines.len() - 1;
+                if line.is_empty() || is_last || is_control_comment {
+                    if let Some(end_index) = end {
+                        Self::sort_part(&mut lines[start_index..end_index]);
+                        end = None;
+                    }
+                    start_index = i + 1;
+                }
+            }
+
+            if controls_this_check {
+                is_disabled = is_off;
+                start_index = i + 1;
+            }
+        }
+
+        let count = warnings.len();
+        for warning in warnings {
+            warning.mark_as_fixed();
+        }
+
+        Some(count)
+    }
+}
+
+impl UnorderedKeyFixer<'_> {
+    fn sort_part(part: &mut [LineEntry]) {
+        // Each slice includes a significant line (with key) and previous comments (if present)
+        let mut slice_start = 0;
+        let mut slices = Vec::with_capacity(part.len());
+        for (i, line) in part.iter().enumerate() {
+            if !line.is_comment() {
+                slices.push(&part[slice_start..=i]);
+                slice_start = i + 1;
+            }
+        }
+
+        slices.sort_by_cached_key(|slice| {
+            // I think, that we should modify get_key() so it will return Option<&str> instead of
+            // Option<String>.
+            slice.last().expect("significant line").get_key()
+        });
+
+        let mut sorted_lines = Vec::with_capacity(part.len());
+        for slice in slices {
+            sorted_lines.extend_from_slice(slice);
+        }
+
+        part.clone_from_slice(sorted_lines.as_slice());
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::common::tests::line_entry;
+
+    fn get_lines(strs: Vec<&str>) -> Vec<LineEntry> {
+        let total = strs.len();
+
+        strs.iter()
+            .enumerate()
+            .map(|(i, &str)| line_entry(i + 1, total, str))
+            .collect()
+    }
+
+    fn get_warnings(lines: &[LineEntry], wrns: Vec<(usize, &str)>) -> Vec<Warning> {
+        wrns.into_iter()
+            .map(|(i, str)| Warning::new(lines[i].clone(), "UnorderedKey", String::from(str)))
+            .collect()
+    }
+
+    fn run_fixer(warnings: &mut [Warning], lines: &mut Vec<LineEntry>) -> Option<usize> {
+        let mut fixer = UnorderedKeyFixer::default();
+        let warning_refs = warnings.iter_mut().collect();
+
+        fixer.fix_warnings(warning_refs, lines)
+    }
+
+    fn assert_lines(lines: &[LineEntry], strs: Vec<&str>) {
+        for (i, &str) in strs.iter().enumerate() {
+            assert_eq!(str, lines[i].raw_string.as_str());
+        }
+    }
+
+    #[test]
+    fn fix_warnings_test() {
+        let mut lines = get_lines(vec!["B=C", "A=B", "D=E", "\n"]);
+        let mut warnings = get_warnings(&lines, vec![(1, "The A key should go before B key")]);
+
+        assert_eq!(Some(1), run_fixer(&mut warnings, &mut lines));
+
+        assert_lines(&lines, vec!["A=B", "B=C", "D=E", "\n"]);
+
+        assert!(warnings[0].is_fixed);
+    }
+
+    #[test]
+    fn many_moves_test() {
+        let mut lines = get_lines(vec!["X=X", "A=A", "D=D", "Z=Z", "Y=Y", "KLM=123", "\n"]);
+        let mut warnings = get_warnings(
+            &lines,
+            vec![
+                (1, "The A key should go before B key"),
+                (4, "The Y key should go before Z key"),
+                (5, "The KLM key should go before Y key"),
+            ],
+        );
+
+        assert_eq!(Some(3), run_fixer(&mut warnings, &mut lines));
+
+        assert_lines(
+            &lines,
+            vec!["A=A", "D=D", "KLM=123", "X=X", "Y=Y", "Z=Z", "\n"],
+        );
+
+        for warning in warnings {
+            assert!(warning.is_fixed);
+        }
+    }
+
+    #[test]
+    fn previous_comments_test() {
+        let mut lines = get_lines(vec![
+            "# X comment 1",
+            "# X comment 2",
+            "X=X",
+            "# A comment",
+            "A=A",
+            "B=B",
+            "Z=Z",
+            "\n",
+        ]);
+        let mut warnings = get_warnings(&lines, vec![(4, "The A key should go before X key")]);
+
+        assert_eq!(Some(1), run_fixer(&mut warnings, &mut lines));
+
+        assert_lines(
+            &lines,
+            vec![
+                "# A comment",
+                "A=A",
+                "B=B",
+                "# X comment 1",
+                "# X comment 2",
+                "X=X",
+                "Z=Z",
+                "\n",
+            ],
+        );
+
+        assert!(warnings[0].is_fixed);
+    }
+
+    #[test]
+    fn several_groups_test() {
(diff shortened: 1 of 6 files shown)


baile320:
Quick question: should this support key sorting within groupings?

For example:
```
# Grouping 1
A=B
C=B
B=D

# Grouping 2
THIS=THAT
Z=X
Q=R
```

after running the fixer could look like:

```
# Grouping 1
A=B
B=D
C=B

# Grouping 2
Q=R
THIS=THAT
Z=X
```

When I run the current fixer, I get:

```
# Grouping 1
A=B
B=D
C=B

Q=R
# Grouping 2
THIS=THAT
Z=X
```
which is not behavior I would expect if I were a user because I think users frequently group common functionality together in env files (all database connections go near each other, all message broker connections near each other, etc). If it is decided to keep the functionality as it currently is, it might be helpful to add explanations in the doc somewhere. Currently, the docs say "You can use blank lines to split lines into groups" (https://github.com/dotenv-linter/dotenv-linter#unordered-key) so that might lead to confusion.

Open to hearing other people's thoughts though.

evgeniy-r:
This behavior is due to the fact that we can handle comments in different ways.

I assumed that the comment is most likely related to the next line with a key (`# Grouping 2` is related to `THIS=THAT`).
So it is handled as a whole.

It can be disscussed and changed if necessary.

evgeniy-r:
And yes, we sort the keys separately in each group.

mgrachev:
@evgeniy-r Thank you for your contribution! 🔥

I need more time to review this PR 🙂

src/fixes/unordered_key.rs
```diff
@@ -0,0 +1,266 @@
...
+    ) -> Option<usize> {
+        // At first we find all groups
+
+        let mut sorted_part_started = false;
+        let mut start = None;
+        let mut end = None;
+
+        let mut sorted_groups = Vec::with_capacity(lines.len());
+
+        for (i, line) in lines.iter().enumerate() {
```
DDtKey, line 48: I think that it is worth changing the approach here, the abundance of `Option` variables complicates the logic and requires unwanted `expect()`:thinking:

Maybe something like this? (quickly written code, may contain an error, just an attempt to convey my idea)
```
        let mut start_part_index = 0;

        for i in 0..lines.len()  {
            match lines.get(i) {
                Some(el) if el.is_empty() => {
                    if start_part_index < i {
                        let part = &mut lines[start_part_index..i];
                        sort_part(part);
                    }
                    start_part_index = i + 1;
                }
                _ => {}
            }
        }
```

This should be consistent with my other comments :slightly_smiling_face:

DDtKey (commented):
You did a great job and thank you for that! :+1: 

I left some related comments, please take a look.

This is by no means a criticism, just considering options :sweat_smile:

DDtKey:
Among other things, I noticed that we work exclusively with `lines` (**without checking warnings!**) and we always sort them.
But we have control comments(#237) and not the entire file should be affected by this `Fixer`, but only really existing `Warning`s.

I understand that the order can break, but this also does not seem to be a suitable solution.
And it's really trouble... 
I also found this in the recently added `EndingBlankLineFixer` (#263)

Perhaps this is beyond the scope of this task at the moment, but we cannot forget about it.

What do you think @dotenv-linter/core @evgeniy-r ?

mgrachev:
> Among other things, I noticed that fixer works exclusively with `lines` (**without checking warnings!**) and always sort full file.
> But we have control comments(#237) and not the entire file should be affected by this `Fixer`, but only really existing `Warning`s.

Good catch! You're absolutely right. All checks and fixes should be disabled using controll comments or the `--skip` argument.

> I understand that the warning `line_number` can break in fix process, but this also does not seem to be a suitable solution.
> And it's really trouble...
> I also found this in the recently added `EndingBlankLineFixer` (#263)

Can you give more information about the problem with `EndingBlankLineFixer` and controll comments?
I've just checked it and it's worked with the controll comment `# dotenv-linter:off EndingBlankLine`.
Or you can open an issue and we will discuss about it.

> One option is of course to check the control comments in the fixer. But this will create duplicate logic.
> At the same time, changing `line_number` in `Warning`s is also not correct, because when displaying results, we must point to the line of error in the original file (but when there are implementations of all fixers, we may not print this at all).
> 
> Perhaps this is beyond the scope of this task at the moment, but we cannot forget about it.

No, I don't think so. We should find a solution before merging this PR 🤔

DDtKey:
> Can you give more information about the problem with `EndingBlankLineFixer` and controll comments?
> I've just checked it and it's worked with the controll comment `# dotenv-linter:off EndingBlankLine`.
> Or you can open an issue and we will discuss about it.

Sorry, I was wrong (typo), I meant `ExtraBlankLineFixer` (#260) and [this review comment](https://github.com/dotenv-linter/dotenv-linter/pull/260#discussion_r471299204)
A little later, I can double-check and create an issue.

> No, I don't think so. We should find a solution before merging this PR thinking

I fully agree.

mgrachev:
> Sorry, I was wrong (typo after GH suggest), I meant `ExtraBlankLineFixer` (#260) and [this review comment](https://github.com/dotenv-linter/dotenv-linter/pull/260#discussion_r471299204)
> A little later, I can double-check and create an issue.

Let me know. Thank you 🙏

evgeniy-r:
I was away, will answer all comments soon.

src/fixes/unordered_key.rs
```diff
@@ -0,0 +1,266 @@
...
+        for warning in warnings {
+            warning.mark_as_fixed();
+        }
+
+        Some(count)
+    }
+}
+
+impl UnorderedKeyFixer<'_> {
+    fn sort_part(lines: &mut [LineEntry], start: usize, end: usize) {
```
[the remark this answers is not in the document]
evgeniy-r: To make sure we are talking about the same thing.

We don't make changes within the slices.
Each slice is a significant line (with key) + previous comments (if present). Most of slices consist of one line.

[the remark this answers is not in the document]
evgeniy-r: Done

src/fixes/unordered_key.rs
```diff
@@ -0,0 +1,266 @@
...
+        warnings: Vec<&mut Warning>,
+        lines: &mut Vec<LineEntry>,
+    ) -> Option<usize> {
+        // At first we find all groups
+
+        let mut sorted_part_started = false;
+        let mut start = None;
+        let mut end = None;
+
+        let mut sorted_groups = Vec::with_capacity(lines.len());
```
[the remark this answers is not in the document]
evgeniy-r: I have removed `sorted_groups`.

src/fixes/unordered_key.rs
evgeniy-r: I didn't quite get the idea (

We don't need `match` in this code snippet, we can use `[]`.

And it doesn't cover all cases (e.g. with comments).

There are 3 variables (`sorted_part_started`, `start`, `end`) and it seems that we need all of them.
A comment may or may not be the start of a sorted group:
```

# This is not
# a sorted group.
# No keys, just comments

```

Also `i` is not always the end of a sorted group:
```

A=B
C=D
# The end of a sorted group is
# the line `C=D`,
# not this one

```

evgeniy-r:
Before discussing control comments, I have a question.
Is this the expected behavior?

File:
```
X=Y
# dotenv-linter:off UnorderedKey
A1=B
A2=B

# dotenv-linter:on UnorderedKey
A=B
```

Linter's output:
```
.env:7 UnorderedKey: The A key should go before the X key

Found 1 problem
```

DDtKey:
> Before discussing control comments, I have a question.
> Is this the expected behavior?

I think there is some logic in this.
Actually, the group is separated in the comment block and it (block) is not surrounded by blank lines
But it's may be not clear.

But from my point of view, this is the expected result, and in this case, the following option looks correct for separation:
```
X=Y
# dotenv-linter:off UnorderedKey
A1=B
A2=B
# dotenv-linter:on UnorderedKey

A=B
```

That is, I would not consider comments attached to specific lines if they are separated from them(key\value line) by a line break :thinking: 

Maybe @mgrachev has a different vision on this?

src/fixes/unordered_key.rs
```diff
@@ -0,0 +1,262 @@
...
+                }
+                if !line.is_comment() {
+                    end = Some(i + 1);
+                    sorted_part_started = true;
+                }
+            }
+        }
+
+        // The last group
+        if sorted_part_started {
```
DDtKey, line 68: What about just add in the body of the `for` cycle additional condition for last group instead of checking after the loop?
For example, in `for`:
```
if line.is_empty() || (i == lines.len() - 1)
```

evgeniy-r: In this case, we need to add more conditions, since the `end` variable has not been assigned a value `lines.len() - 1`.

DDtKey: Yeah, you're right, sorry :slightly_smiling_face:

src/fixes/unordered_key.rs
DDtKey: In fact, 3 variables seem redundant to me. I apologize that the code snippet was not clear.

Suppose in your version, the _start index_ is always shifted by 1 after finding an empty string (in next iteration). We may not use `Option` here.

In the case of `sorted_part_started`, I see a clear correlation with the `end`.

How do you like this option? He must pass all tests.

```
        let mut start = 0;
        let mut end = None;

        for i in 0..lines.len() {
            let line = &lines[i];
            let is_last_line = i == lines.len() - 1;

            if line.is_empty() || is_last_line {
                if let Some(end_index) = end {
                    Self::sort_part(
                        &mut lines[start..end_index],
                    );
                    end = None;
                }
                start = i + 1;
            } else if !line.is_comment() {
                end = Some(i + 1);
            }
        }
```

Perhaps this can be done even more expressively. But this is still the main idea.

Also, this option saves us from the need to check the last group outside the loop.

src/fixes/unordered_key.rs
DDtKey: Please see [this comment](https://github.com/dotenv-linter/dotenv-linter/pull/261#discussion_r481022360), it solves this issue

DDtKey: Please see [this comment](https://github.com/dotenv-linter/dotenv-linter/pull/261#discussion_r481022360), it solves this issue

src/fixes/unordered_key.rs
DDtKey: I also noticed that we seem to be missing a test for something where the last line is not empty, but here we will get two `Warning`s (`UnorderedKey` and `EndingBlankLine`)

For example:
```dotenv
       # I comment\n\
        I=123\n\
        \n\
        B1=1\n\
        A1=1",
    );
```
Maybe add an integration test?


Therefore, to check the last line fully correct, we can slightly change the end condition:
```
        let mut start = 0;
        let mut end = None;

        for i in 0..lines.len() {
            let line = &lines[i];
            let is_last_line = i == lines.len() - 1;

            if !line.is_empty_or_comment(){
                end = Some(i + 1);
            }

            if line.is_empty() || is_last_line {
                if let Some(end_index) = end {
                    Self::sort_part(
                        &mut lines[start..end_index],
                    );
                    end = None;
                }
                start = i + 1;
            }
        }
```

src/fixes/unordered_key.rs
DDtKey: > In this case, we need to add more conditions, since the `end` variable has not been assigned a value `lines.len() - 1`.

Current tests cannot cover this condition anyway, they always have `\n` at the end. Please see also [comment](https://github.com/dotenv-linter/dotenv-linter/pull/261#discussion_r481040776)

DDtKey:
> > Sorry, I was wrong (typo after GH suggest), I meant `ExtraBlankLineFixer` (#260) and [this review comment](https://github.com/dotenv-linter/dotenv-linter/pull/260#discussion_r471299204)
> > A little later, I can double-check and create an issue.
> 
> Let me know. Thank you

I created an issue #278 with examples :slightly_smiling_face:

mgrachev:
> At the same time, changing `line_number` in `Warning`s is also not correct, because when displaying results, we must point to the line of error in the original file (but when there are implementations of all fixers, we may not print this at all).

I think we should remove the logic for displaying `Fixed/Unfixed Warnings` after we merge this PR.

evgeniy-r:
> 
> I think there is some logic in this.
> Actually, the group is separated in the comment block and it (block) is not surrounded by blank lines. But it's may be not clear.
> 

So, the answer is "yes" (that is the desired behavior)? :slightly_smiling_face:

mgrachev:
> So, the answer is "yes" (that is the desired behavior)? 🙂

Yes, the `UnorderedKey` check works as expected.

src/fixes/unordered_key.rs
evgeniy-r: This solution is simpler, thank you.
I have used it (I have rename `start` to `start_index`) and have added the test (unit kind, I think this is OK).

src/fixes/unordered_key.rs
evgeniy-r: Done

evgeniy-r:
> Yes, the UnorderedKey check works as expected.

I plan to submit a PR tomorrow (control comments).

evgeniy-r:
I found an interesting kind of issues:

Before fixing the order:
```
BB=1
# dotenv-linter:off LowercaseKey
Aa=B
X=X
# dotenv-linter:on LowercaseKey

```

After fixing:
```
Aa=B
BB=1
# dotenv-linter:off LowercaseKey
X=X
# dotenv-linter:on LowercaseKey

```
And we will get the incorrect file.
If we fix the `Aa` key somehow, it will not be the best solution either, because a user doesn't want fix the lower case.

Therefore, I believe that we should split the groups by control comments, as well as by blank lines.

I made appropriate changes.

evgeniy-r:
I moved changes in checker to the separate PR: https://github.com/dotenv-linter/dotenv-linter/pull/281

src/fixes/unordered_key.rs
```diff
@@ -0,0 +1,724 @@
...
+                is_off = comment.is_disabled();
+            }
+
+            if !is_disabled {
+                if !line.is_empty_or_comment() {
+                    end.replace(i + 1);
+                }
+
+                let is_last = i == lines.len() - 1;
+                if line.is_empty() || is_last || is_control_comment {
```
mgrachev, line 66:
```suggestion
                if line.is_empty() || line.is_last_line() || is_control_comment {
```
Why don't you use the `is_last_line` method?

src/fixes/unordered_key.rs
```diff
@@ -0,0 +1,724 @@
...
+            if !line.is_comment() {
+                slices.push(&part[slice_start..=i]);
+                slice_start = i + 1;
+            }
+        }
+
+        slices.sort_by_cached_key(|slice| {
+            // I think, that we should modify get_key() so it will return Option<&str> instead of
+            // Option<String>.
+            slice.last().expect("significant line").get_key()
```
mgrachev, line 105:
```suggestion
            slice.last()?.get_key()
```

src/fixes/unordered_key.rs
```diff
@@ -0,0 +1,724 @@
...
+mod tests {
+    use super::*;
+    use crate::common::tests::line_entry;
+
+    fn get_lines(strs: Vec<&str>) -> Vec<LineEntry> {
+        let total = strs.len();
+
+        strs.iter()
+            .enumerate()
+            .map(|(i, &str)| line_entry(i + 1, total, str))
```
mgrachev, line 127:
```suggestion
            .map(|(i, &line)| line_entry(i + 1, total, line))
```
Please don't use `str` as a variable name, because Rust has the same type's name - [str](https://doc.rust-lang.org/std/primitive.str.html).

src/fixes/unordered_key.rs
```diff
@@ -0,0 +1,724 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::common::tests::line_entry;
+
+    fn get_lines(strs: Vec<&str>) -> Vec<LineEntry> {
+        let total = strs.len();
+
+        strs.iter()
```
mgrachev, line 125:
```suggestion
    fn get_lines(lines: Vec<&str>) -> Vec<LineEntry> {
        let total = lines.len();

        lines.iter()
```

src/fixes/unordered_key.rs
```diff
@@ -0,0 +1,724 @@
...
+            .enumerate()
+            .map(|(i, &str)| line_entry(i + 1, total, str))
+            .collect()
+    }
+
+    fn get_warnings(lines: &[LineEntry], wrns: Vec<(usize, &str)>) -> Vec<Warning> {
+        wrns.into_iter()
+            .map(|(i, str)| Warning::new(lines[i].clone(), "UnorderedKey", String::from(str)))
```
mgrachev, line 133:
```suggestion
    fn get_warnings(lines: &[LineEntry], warnings: Vec<(usize, &str)>) -> Vec<Warning> {
        warnings.into_iter()
            .map(|(i, line)| Warning::new(lines[i].clone(), "UnorderedKey", String::from(line)))
```

src/fixes/unordered_key.rs
```diff
@@ -0,0 +1,724 @@
...
+    fn run_fixer(warnings: &mut [Warning], lines: &mut Vec<LineEntry>) -> Option<usize> {
+        let mut fixer = UnorderedKeyFixer::default();
+        let warning_refs = warnings.iter_mut().collect();
+
+        fixer.fix_warnings(warning_refs, lines)
+    }
+
+    fn assert_lines(lines: &[LineEntry], strs: Vec<&str>) {
+        for (i, &str) in strs.iter().enumerate() {
+            assert_eq!(str, lines[i].raw_string.as_str());
```
mgrachev, line 146:
```suggestion
    fn assert_lines(result: &[LineEntry], lines: Vec<&str>) {
        for (i, &line) in lines.iter().enumerate() {
            assert_eq!(line, result[i].raw_string.as_str());
```

tests/common/mod.rs
```diff
@@ -128,6 +128,7 @@ impl TestDir {
     }
 
     /// Run the default CLI binary, with "-f", in this TestDir and check it fails.
+    #[allow(dead_code)]
```
mgrachev, line 131: Please remove this method.

tests/fixes/mod.rs
```diff
@@ -54,30 +54,96 @@ fn key_without_value() {
 }
 
 #[test]
-fn unfixed_warnings() {
+fn unordered_key() {
```
mgrachev, line 57: Please move this test to the `unordered_key.rs` file.

tests/fixes/mod.rs
```diff
@@ -54,30 +54,96 @@ fn key_without_value() {
...
+        A1=1\n\
+        B1=1\n\
+        \n\
+        # end comment\n",
     );
 
     testdir.close();
 }
 
+// We need to figure out how to test unfixed warnings now when all the fixers have been implemented
```
mgrachev, line 119: Please remove this test.

evgeniy-r: Done

tests/fixes/mod.rs
evgeniy-r: Done

tests/common/mod.rs
evgeniy-r: Done

src/fixes/unordered_key.rs
evgeniy-r: Done

src/fixes/unordered_key.rs
evgeniy-r: Done

src/fixes/unordered_key.rs
evgeniy-r: Done

src/fixes/unordered_key.rs
evgeniy-r: Done

src/fixes/unordered_key.rs
evgeniy-r: Done

src/fixes/unordered_key.rs
evgeniy-r: Now we don't delete lines (just mark them), so this is OK.

evgeniy-r: Done

mgrachev (approved):

PR #267, opened 2020-08-19
Update logic for IncorrectDelimiterChecker

baile320:
I ran into issues with IncorrectDelimiterChecker: it thought, for example, "*FOO=BAR" had an incorrect delimiter (this type of error will ultimately fixed by LeadingCharacterFixer, see: https://github.com/dotenv-linter/dotenv-linter/pull/259). IMO, that is not a delimiter at all, since delimiters should occur /between/ other characters, so I updated the logic of this checker and updated some of the tests & added new ones.

#### ✔ Checklist:


src/common/remove_all_invalid_leading_chars.rs
```diff
@@ -0,0 +1,18 @@
+pub fn remove_all_invalid_leading_chars(string: &str) -> String {
+    let mut cleaned_string = string;
+
+    while !cleaned_string.starts_with(|c: char| c.is_alphabetic() || c == '_') {
```
gillespiecd, line 4: I would recommend using the `skip_while` iterator, prevents the need to copy after each skipped char.

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.skip_while

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=acd50e88e87a367055f4da35e46febda

baile320: Thank you! I did not know about that function.

src/common/remove_all_invalid_leading_chars.rs
```diff
@@ -0,0 +1,15 @@
+pub fn remove_all_invalid_leading_chars(string: &str) -> String {
+    string
+        .chars()
+        .skip_while(|c| !(c.is_alphabetic() || *c == '_'))
```
mgrachev, line 4:
```suggestion
        .skip_while(|&c| !(c.is_alphabetic() || c == '_'))
```

src/common.rs
```diff
@@ -1,9 +1,11 @@
 mod file_entry;
 mod line_entry;
+mod remove_all_invalid_leading_chars;
 mod warning;
 
 pub use file_entry::FileEntry;
 pub use line_entry::LineEntry;
+pub use remove_all_invalid_leading_chars::remove_all_invalid_leading_chars;
```
mgrachev, line 8: It would be better to leave this function in the `common.rs` file.

mgrachev (commented):
@baile320 Thanks you for your help! 👍

I have left some comments. Please look at them 👀

src/common/remove_all_invalid_leading_chars.rs
```diff
@@ -0,0 +1,15 @@
+pub fn remove_all_invalid_leading_chars(string: &str) -> String {
```
DDtKey, line 1: Small suggestion, I think "all" excess here, because the name is still clear but shorter
```suggestion
pub fn remove_invalid_leading_chars(string: &str) -> String {
```

baile320:
Thanks everyone for your patience! I have made the requested changes.

src/common.rs
```diff
@@ -1,11 +1,25 @@
...
 pub use line_entry::LineEntry;
-pub use remove_all_invalid_leading_chars::remove_all_invalid_leading_chars;
 pub use warning::Warning;
 
 pub const LF: &str = "\n";
+
+pub fn remove_invalid_leading_chars(string: &str) -> String {
+    string
+        .chars()
+        .skip_while(|c| !(c.is_alphabetic() || *c == '_'))
```
mgrachev:
```suggestion
        .skip_while(|&c| !(c.is_alphabetic() || c == '_'))
```

mgrachev (approved):
LGTM 👍

mgrachev:

🙏 If it’s not difficult for you, please support the project - click on the star on GitHub ⭐️

Issue #243, opened 2020-08-10
Add fixer: DuplicatedKeyFixer


DDtKey:
Implement `trait Fix` to fix warnings caused by `DuplicatedKeyChecker`.
And add this `Fix` to the `fixlist` in [fixes.rs](https://github.com/dotenv-linter/dotenv-linter/blob/cf176993360215596c079873725f4a4b7a48f5b0/src/fixes.rs#L32)

I think this fixer should remove duplicate keys. 
For example:
```
PROFILE_ENV=development
PROFILE_ENV=test
```
After running `DuplicatedKeyFixer` it should look like this:
```
PROFILE_ENV=development
```

Implementation example: [LowercaseKeyFixer](https://github.com/dotenv-linter/dotenv-linter/blob/master/src/fixes/lowercase_key.rs)

mgrachev:
> I think this fixer should remove duplicate keys.

What do you think about commenting a duplicated key instead of removing it?

DDtKey:
> > I think this fixer should remove duplicate keys.
> 
> What do you think about commenting the duplicated key instead of removing it?

Yes, maybe this is a good approach :+1:  
At least until we have backups.

PR #270, opened 2020-08-20
solve #243: add DuplicatedKeyFixer

utter-step:
Solves #243

#### ✔ Checklist:

### DIFF
--- a/src/fixes.rs
+++ b/src/fixes.rs
@@ -1,5 +1,6 @@
 use crate::common::*;
 
+mod duplicated_key;
 mod ending_blank_line;
 mod key_without_value;
 mod lowercase_key;
@@ -17,7 +18,10 @@ trait Fix {
     ) -> Option<usize> {
         let mut count: usize = 0;
         for warning in warnings {
-            let line = lines.get_mut(warning.line_number() - 1)?;
+            let line = lines
+                .get_mut(warning.line_number() - 1)
+                .expect("warning references inexistent line");
+
             if self.fix_line(line).is_some() {
                 warning.mark_as_fixed();
                 count += 1;
@@ -45,6 +49,7 @@ fn fixlist() -> Vec<Box<dyn Fix>> {
         // Then we should run the fixers that handle the line entry collection at whole.
         // And at the end we should run the fixer for ExtraBlankLine check (because the previous
         // fixers can create additional extra blank lines).
+        Box::new(duplicated_key::DuplicatedKeyFixer::default()),
         Box::new(ending_blank_line::EndingBlankLineFixer::default()),
     ]
 }
--- a/src/fixes.rs
+++ b/src/fixes.rs
@@ -155,30 +155,6 @@ mod tests {
         assert!(!warnings[0].is_fixed);
     }
 
-    #[test]
-    fn run_when_lines_do_not_fit_numbers_test() {
-        let mut lines = vec![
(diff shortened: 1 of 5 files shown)


src/fixes.rs
```diff
@@ -17,7 +18,10 @@ trait Fix {
     ) -> Option<usize> {
         let mut count: usize = 0;
         for warning in warnings {
-            let line = lines.get_mut(warning.line_number() - 1)?;
+            let line = lines
+                .get_mut(warning.line_number() - 1)
+                .expect("warning references inexistent line");
```
utter-step, line 23: I guess that this should be hard error, instead of silent pass?

Also, I'm feeling slightly uncomfortable about relating on the fact that the lines count/order is preserved after previous fixers... For example, fixer for UnorderedKey can mangle the lines order here. Wouldn't it be better for Warning to store &mut reference to the line instead of it's clone? Or maybe we should use `find` based on `line_number` here instead of `get_mut`?

src/fixes.rs
```diff
@@ -151,30 +156,6 @@ mod tests {
...
-            Warning::new(
-                lines[1].clone(),
-                "LowercaseKey",
-                String::from("The c key should be in uppercase"),
-            ),
-        ];
-
-        assert_eq!(0, run(&mut warnings, &mut lines));
-        assert!(!warnings[0].is_fixed);
-    }
```
utter-step, line 176: I'm not sure it's OK to delete this test, but this is linked to the question above

src/fixes/duplicated_key.rs
```diff
@@ -0,0 +1,152 @@
...
+        lines: &mut Vec<LineEntry>,
+    ) -> Option<usize> {
+        let len = warnings.len();
+
+        for warning in warnings {
+            let line = lines
+                .get_mut(warning.line_number() - 1)
+                .expect("warning references inexistent line");
+
+            line.raw_string = format!("#{}", line.trimmed_string());
```
baile320, line 33: I'm wondering if it's necessary to do ```.trimmed_string``` here - presumably if there is extra whitespace on a line, that should be fixed by another Fixer? (I'm not sure about this, so if there's a good reason I'd like to hear why)

src/fixes/duplicated_key.rs
```diff
@@ -0,0 +1,152 @@
...
+    fn name(&self) -> &str {
+        self.name
+    }
+
+    fn fix_warnings(
+        &self,
+        warnings: Vec<&mut Warning>,
+        lines: &mut Vec<LineEntry>,
+    ) -> Option<usize> {
+        let len = warnings.len();
```
baile320, line 26: can we define this closer to where it's used (or maybe just not declare it at all, ```Some(warnings.len())``` seems readable too

mgrachev:
> > I guess we should run not only fixer, but also checker.
> 
> There are two options:
> 
> 1. We can implement the `fix_warnings` function, not the `fix_line`, and comment out all duplicated keys, not just the ones found during the initial check. Yes, there will be a slight duplication of code with the checker, but it seems to me that such an implementation is simpler and more reliable.
> 2. We can run the checker again somehow to find new warnings.
> 
> I prefer the first option but would like to discuss different approaches.

I'd think about it after we do all the fixer tasks 🤔

mgrachev (approved):

mgrachev:
@utter-step Thank you for your help! 👍

PR #272, opened 2020-08-21
Implement Backup Feature

baile320:
Hey All, I had some questions/comments about this PR. The basic functionality is in the PR to create a copy of a file via the -c/--copy flag, though.

Questions:
1) I just used a Unix epoch time timestamp. We could add a more nicely formatted one, but it seems likely we'd need to add a crate (I'm not sure?), and I don't know if that's wanted or needed. The Unix timestamps still allow the copies to easily be sorted by most recent.

2) Should copying the file be automatic any time there are fixes?
    1) If it should be automatic, should there be an opt-out flag? If so, what flag? -n//--no-copy?
    2) If users should always specify a flag to create a copy, should we prompt them that their file will get overwritten?


src/lib.rs
```diff
@@ -52,6 +52,13 @@ pub fn run(args: &clap::ArgMatches, current_dir: &PathBuf) -> Result<Vec<Warning
 
         let mut lines = get_line_entries(&fe, strs);
 
+        // create copy of file if user requests it
```
baile320, line 55: probably, we should only create a backup if ```fixes::run(&mut result, &mut lines) > 0```

src/fs_utils.rs
```diff
@@ -34,6 +36,7 @@ pub fn get_relative_path(target_path: &PathBuf, base_path: &PathBuf) -> Option<P
     Some(relative_path)
 }
 
+// TODO: add the notification if user doesn't use -c/--copy flag
```
baile320, line 39: we need to do something about this todo

baile320:
Just an FYI - I haven't forgotten about this, just been busy. I think I'll have time do make the changes later today.

src/fs_utils.rs
```diff
@@ -47,6 +50,21 @@ pub fn write_file(path: &PathBuf, lines: Vec<LineEntry>) -> io::Result<()> {
...
+pub fn copy_file(fe: &FileEntry, lines: Vec<LineEntry>) -> Result<PathBuf, Box<dyn Error>> {
+    let timestamp = SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)?
+        .as_secs();
+    let new_filepath = PathBuf::from(format!(
+        "{file_entry}_{timestamp}",
+        file_entry = fe,
+        timestamp = timestamp
+    ));
+    match write_file(&new_filepath, lines) {
```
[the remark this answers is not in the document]
baile320: done

src/lib.rs
baile320: done

baile320:
I believe this should be ready for review now. Changes made since discussion:

1. Renamed feature from Copy to Backup (changed arg to "no-backup", --no-backup, and no short name)
2. Using fs::copy instead of fs::write to create the file.
3. Inform users of the file created after creating it.
4. Only back up file if the .env file is being modified
5. Added the backup files to .gitignore as well (*.env_*)

baile320:
I'm not sure why the ```CI / Test and coverage / Linux (pull_request) Failing after 45s — Test and coverage / Linux``` is failing, is this something I need to look into? I see a couple other PRs where it failed as well.

evgeniy-r:
> I'm not sure why the CI / Test and coverage / Linux (pull_request) Failing after 45s — Test and coverage / Linux is failing, is this something I need to look into?

This is due to a change in the output:
https://github.com/dotenv-linter/dotenv-linter/pull/272/checks?check_run_id=1047857184#step:9:210

src/fs_utils.rs
```diff
@@ -47,6 +49,22 @@ pub fn write_file(path: &PathBuf, lines: Vec<LineEntry>) -> io::Result<()> {
     Ok(())
 }
 
+pub fn backup_file(path: &PathBuf) -> Result<PathBuf, Box<dyn Error>> {
+    let timestamp = SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)?
+        .as_secs();
+    let orig_path = path.to_str().unwrap();
```
mgrachev, line 56: It would be great not to use the `unwrap` method here.

src/fs_utils.rs
```diff
@@ -168,4 +186,51 @@ mod tests {
...
+                number: 3,
+                file: fe.clone(),
+                raw_string: String::from("\n"),
+            },
+        ];
+
+        if write_file(&fe.path, lines).is_ok() {
+            match backup_file(&fe.path) {
+                Ok(path) => {
+                    assert_eq!(
```
mgrachev, line 223: Please also check that `path` is not equal to `fe.path`.

mgrachev:
> This is due to a change in the output:
> https://github.com/dotenv-linter/dotenv-linter/pull/272/checks?check_run_id=1047857184#step:9:210

Please also update all integration tests and write another one to cover the `--no-backup` flag.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### 🚀 Added
+- Add backup/no-backup flag for fixers [#272](https://github.com/dotenv-linter/dotenv-linter/pull/272) ([@baile320](https://github.com/baile320))
```
DDtKey, line 9:
```suggestion
- Add backup feature and `--no-backup` flag for fixers [#272](https://github.com/dotenv-linter/dotenv-linter/pull/272) ([@baile320](https://github.com/baile320))
```

src/lib.rs
```diff
@@ -52,8 +52,17 @@ pub fn run(args: &clap::ArgMatches, current_dir: &PathBuf) -> Result<Vec<Warning
 
         let mut lines = get_line_entries(&fe, strs);
 
+        // run fixers & write results to file
         let mut result = checks::run(&lines, &skip_checks);
         if is_fix && fixes::run(&mut result, &mut lines) > 0 {
+            // create backup copy unless user specifies not to
+            let should_backup = !args.is_present("no-backup");
+            if is_fix && should_backup {
```
DDtKey, line 60: `is_fix` seems extra, because code already in `if` snippet :slightly_smiling_face: 
```suggestion
            if should_backup {
```

DDtKey (commented):
Thank you for you contribution! :rocket: 
I left some comments, please take a look them.

baile320:
> Please also update all integration tests and write another one to cover the `--no-backup` flag.

Still need to complete this task, will get to it later today and post a reply when completed

baile320:
hmm... i'm not sure how much detail I should be checking when a backup is created in an integration test. since the file is created with a timestamp, i'm not sure how to know when testing what that timestamp would be? I could just check for a wildcard file or something if that's necessary? I feel a little stuck on testing these fs components.

but in the unit test i already do make sure the filepath & backup filepath are different, and that the backup filepath has the original environment.

src/fs_utils.rs
```diff
@@ -47,6 +49,26 @@ pub fn write_file(path: &PathBuf, lines: Vec<LineEntry>) -> io::Result<()> {
...
+            orig_path = orig_path,
+            timestamp = timestamp
+        ));
+    }
+
+    match copy(path, &new_filepath) {
+        Ok(_) => Ok(new_filepath),
+        Err(e) => Err(Box::new(e)),
+    }
+}
```
mgrachev, line 70:
```suggestion
pub fn backup_file(fe: &FileEntry) -> Result<PathBuf, Box<dyn Error>> {
    let timestamp = SystemTime::now()
        .duration_since(SystemTime::UNIX_EPOCH)?
        .as_secs();

    let mut new_path = fe.path.to_owned();
    new_path.set_file_name(format!("{}_{}", &fe.file_name, timestamp));

    match copy(&fe.path, &new_path) {
        Ok(_) => Ok(new_path),
        Err(e) => Err(Box::new(e)),
    }
}
```

I think you can get the file name from the `FileEntry` object and then add a timestamp to the new file path.
What do you think about this way?

tests/common/mod.rs
```diff
@@ -132,13 +132,36 @@ impl TestDir {
...
-            .args(&["-f"])
+            .args(&["-f", "--no-backup"])
             .assert()
             .failure()
             .code(1)
             .stdout(expected_output);
     }
 
+    /// Run the default CLI binary, with "-f", in this TestDir and checks a backup file is created
+    pub fn test_command_fix_backs_up_files(self) {
```
mgrachev, line 143: I think creating a function to run in only one test is a bit overkill 🤔

mgrachev:
> hmm... i'm not sure how much detail I should be checking when a backup is created in an integration test. since the file is created with a timestamp, i'm not sure how to know when testing what that timestamp would be? I could just check for a wildcard file or something if that's necessary? I feel a little stuck on testing these fs components.
> 
> but in the unit test i already do make sure the filepath & backup filepath are different, and that the backup filepath has the original environment.

In my opinion, in the intragration test you should create a file with some problems, for example: `foo=bar`.
Then, you should run `dotenv-linter` with the `-f` flag and check that:
1. The content of the `.env` file doesn't equal `foo=bar`;
2. The content of the `backup` file equal `foo=bar`;

I think that is all.

baile320:
Hm. I think having a separate function for a single test is probably overkill as well, but I think I need a testing function that has some different functionality than the current ones. 

I think I need a test function that will let me 1) supply arguments 2) not close/delete TestDir when done and 3) doesn't require me to supply the expected output. The reasons for those are 1) I changed the other tests to have ```-f --no-backup``` (I think this is OK, it preserves the functionality of the tests that existed prior to adding this feature),  2) I need to be able to the examine the TestDir for the backup file after running, and 3) I don't know what the expected output will be because we're generating a dynamic timestamp every time we run, and I don't see an easy way of mocking that value.

I'm open to suggestions on any of these points, especially the last one.

tests/common/mod.rs
baile320: i think you're right, but see my comment below about the reasons why I made one (also, I updated the function based on your comment below)

src/fs_utils.rs
baile320: I think that is much clearer and made that change. Thank you.

tests/flags/backup.rs
```diff
@@ -0,0 +1,34 @@
...
+                return current_path == testfile.as_str();
+            }
+            false
+        })
+        .collect::<Vec<fs::DirEntry>>();
+    let backup_contents =
+        String::from_utf8_lossy(&fs::read(backup_path[0].path()).expect("read file")).into_owned();
+    assert_eq!(backup_contents, "foo=bar\n");
+
+    testdir.close()
```
mgrachev, line 33:
```suggestion
    let testdir = TestDir::new();
    let content = "foo=bar\n";
    let testfile = testdir.create_testfile(".env", content);
    let args = &[testfile.as_str(), "-f"];

    testdir.test_command_success_with_args_without_closing(args);

    let backup_file = fs::read_dir(&testdir.as_str())
        .expect("read dir")
        .filter_map(|e| e.ok())
        .filter(|e| e.path().as_os_str() != testfile.as_str())
        .last()
        .expect("get backup file");

    let backup_content = fs::read_to_string(backup_file.path()).expect("read backup file");

    assert_eq!(backup_content, content);
    assert_eq!(testfile.contents(), "FOO=bar\n");

    testdir.close()
```
I think you can do it a little easier 🤔 What do you think?

mgrachev:
> Hm. I think having a separate function for a single test is probably overkill as well, but I think I need a testing function that has some different functionality than the current ones.
> 
> I think I need a test function that will let me
> 
> 1. supply arguments
> 2. not close/delete TestDir when done, and doesn't take ownership
> 3. doesn't require me to supply the expected output.
> 
> The reasons for those are:
> 
> 1. I changed the other tests to have `-f --no-backup` (I think this is OK, it basically acts to preserve the functionality of the tests that existed prior to adding this feature),
> 2. I need to be able to the examine the TestDir for the backup file after running
> 3. I don't know what the expected output will be because we're generating a dynamic timestamp every time we run, and I don't see an easy way of mocking that value.
> 
> I'm open to suggestions on any of these points, especially the last one.

👌 Let it be 🙂

tests/flags/backup.rs
baile320: i think that looks much nicer! i made the change.

mgrachev (approved):

mgrachev:

PR #279, opened 2020-09-02
Fix a bug with ExtraBlankLineFixer and control comments

### LINKED ISSUE #278: ExtraBlankLineFixer with control comments work incorrectly
I found a bug when we run dotenv-linter with control comments(#237) and `--fix` flag (now specific for ExtraBlankLineFixer #260).
This is because the fixer works on all line collection without matching `Warning` line number or re-checking the control comments.

<details>
<summary> Example </summary>

```dotenv
DEBUG_HTTP=true

# dotenv-linter:off ExtraBlankLine



# dotenv-linter:on ExtraBlankLine



DB_NAME=development
```

After run `dotenv-linter -f .env` we will receive next result:

```dotenv
DEBUG_HTTP=true

# dotenv-linter:off ExtraBlankLine

# dotenv-linter:on ExtraBlankLine

DB_NAME=development
```

With output:
```
Fixed warnings:
../../workdir/test/test.env:9 ExtraBlankLine: Extra blank line detected
../../workdir/test/test.env:10 ExtraBlankLine: Extra blank line detected
```
</details>

Ie also fixed warnings in the comment block.

**Simple solution:** in such cases, check control comments additionally.
Perhaps it makes sense to move some of the code into a module.

Or we need to **use a different approach** in correcting and matching, but after other fixers, the line order may change, and changing the line number in `Warning` does not look the best way - this way we will not be able to display the numbers of the original file (unless everything is fixed and the output is needed)
Or, as has already been suggested, run them one after another. But all this **requires more changes.**

Now it is also relevant for #261.

Any suggestions? :thinking:

mgrachev:
#### ✔ Checklist:

- [x] Tests for the changes have been added (for bug fixes / features).

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -39,6 +39,7 @@ pub(crate) mod tests {
                 total_lines,
             },
             raw_string: String::from("\n"),
+            is_deleted: false,
         }
     }
 
@@ -52,6 +53,7 @@ pub(crate) mod tests {
                 total_lines,
             },
             raw_string: String::from(raw_string),
+            is_deleted: false,
         }
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add installation CI test for Windows (via `install.sh`) [#235](https://github.com/dotenv-linter/dotenv-linter/pull/235) ([@DDtKey](https://github.com/DDtKey))
 
 ### 🔧 Changed
+- Fix a bug with `ExtraBlankLineFixer` and control comments [#279](https://github.com/dotenv-linter/dotenv-linter/pull/279) ([@mgrachev](https://github.com/mgrachev))
 - Move logic for creating `LineEntry` for tests to `common` module [#280](https://github.com/dotenv-linter/dotenv-linter/pull/280) ([@mgrachev](https://github.com/mgrachev))
 - Simplify UnorderedKeyChecker [#277](https://github.com/dotenv-linter/dotenv-linter/pull/277) ([@mgrachev](https://github.com/mgrachev))
 - Partition fixed/unfixed warnings [#275](https://github.com/dotenv-linter/dotenv-linter/pull/275) ([@gillespiecd](https://github.com/gillespiecd))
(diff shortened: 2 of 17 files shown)


src/common.rs
```diff
@@ -24,3 +24,36 @@ fn remove_invalid_leading_chars_test() {
...
+                file_name: ".env".to_string(),
+                total_lines,
+            },
+            raw_string: String::from("\n"),
+            is_deleted: false,
+        }
+    }
+
+    #[allow(dead_code)]
+    pub fn line_entry(number: usize, total_lines: usize, str: &str) -> LineEntry {
```
baile320, line 47: I have a couple questions/suggestions for my own knowledge here:

1. Is there a reason to use a ``line_entry()``` function instead of a ```LineEntry::new()``` constructor?
2. I don't think it's bad to use ```str``` as an argument (it's not a reserved keyword or anything to my knowledge, but maybe confusing since it's also kind of related to a type?), but maybe the final argument could be called ```raw_string``` since that is the actual struct property's name?

mgrachev: > I have a couple questions/suggestions for my own knowledge here:
> 
> 1. Is there a reason to use a `line_entry()` function instead of a `LineEntry::new()` constructor? I can think of a reason to use a constructor (keep associated functionality of struct in the same place), but I'm wondering if there's a reason to leave it as a standalone function?

We don't need to have the constructor function `LineEntry::new()`, because we create `LineEntry` only in two places.
The `line_entry()` function is needed only for tests and it localed inside the `tests` module.

> 2. I don't think it's bad to use `str` as an argument (it's not a reserved keyword or anything to my knowledge, but maybe confusing since it's also kind of related to a type?), but maybe the final argument could be called `raw_string` since that is the actual struct property's name?

I agree 👍

DDtKey (commented):
Good job, thank you! :+1: 

The only thing, as it seems to me, refactoring the logic of creating `LineEntry` complicates this commit (many files) and perhaps it would be clearer in separate PR (not necessary, it's just loosely coupled changes)

And also, please, add this commit to `CHANGLOG.md` :slightly_smiling_face:

mgrachev:
> The only thing, as it seems to me, refactoring the logic of creating `LineEntry` complicates this commit (many files) and perhaps it would be clearer in separate PR (not necessary, it's just loosely coupled changes)

Done https://github.com/dotenv-linter/dotenv-linter/pull/280

DDtKey (approved):

PR #281, opened 2020-09-11
Group lines by control comments

evgeniy-r:
I changed the `UnorderedKey` check's logic.

Now it also splits lines into sortable groups by control comments.

This is done to make the linter logic more predictable.

#### Before this PR:
File:
```
BB=1
# dotenv-linter:off LowercaseKey
Aa=B
X=X

```
There is one warning (line 3).

After fixing we expect something like this:
```
Aa=B
BB=1
# dotenv-linter:off LowercaseKey
X=X

```

And we will get the incorrect file after fixing.
If we fix the `Aa` key somehow, it will not be the best solution either, because a user obviously doesn't want fix the lower case.

#### After this PR:
File:
```
BB=1
# dotenv-linter:off LowercaseKey
Aa=B
X=X

```
There is no warnings.

#### ✔ Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -449,6 +449,19 @@ FOO=BAR
 BAR=FOO
 ```
 
+Control comments also split lines (this is done to make the linter logic more predictable):
+
+```env
+❌ Wrong
+FOO=BAR
+BAR=FOO
+
+✅ Correct 
+FOO=BAR
+# dotenv-linter:off LowercaseKey
+bar=FOO
+```
+
 ## 🤝 Contributing
 
 If you've ever wanted to contribute to open source, now you have a great opportunity:
--- a/README.md
+++ b/README.md
@@ -449,7 +449,7 @@ FOO=BAR
 BAR=FOO
 ```
 
-Control comments also split lines (this is done to make the linter logic more predictable):
+Control comments also split lines (this is done to make the linter logic more predictable, will be available in [v2.2.0](https://github.com/dotenv-linter/dotenv-linter/issues/238)):
 
 ```env
 ❌ Wrong
--- a/docs/checks/unordered_key.md
+++ b/docs/checks/unordered_key.md
@@ -25,7 +25,7 @@ FOO=BAR
 BAR=FOO
 ```
 
-Control comments also split lines (this is done to make the linter logic more predictable):
+Control comments also split lines (this is done to make the linter logic more predictable, will be available in [v2.2.0](https://github.com/dotenv-linter/dotenv-linter/issues/238)):
 
 ```env
 ❌ Wrong
--- a/docs/checks/unordered_key.md
+++ b/docs/checks/unordered_key.md
@@ -24,3 +24,16 @@ FOO=BAR
 
 BAR=FOO
 ```
+
+Control comments also split lines (this is done to make the linter logic more predictable):
+
+```env
+❌ Wrong
+FOO=BAR
+BAR=FOO
+
+✅ Correct 
+FOO=BAR
+# dotenv-linter:off LowercaseKey
+bar=FOO
+```
--- a/src/checks.rs
+++ b/src/checks.rs
@@ -57,7 +57,7 @@ pub fn run(lines: &[LineEntry], skip_checks: &[&str]) -> Vec<Warning> {
     let mut warnings: Vec<Warning> = Vec::new();
 
     for line in lines {
-        if let Some(comment) = line.get_comment() {
+        if let Some(comment) = line.get_control_comment() {
             if comment.is_disabled() {
                 // Disable checks from a comment using the dotenv-linter:off flag
                 disabled_checks.extend(comment.checks);
--- a/src/checks.rs
+++ b/src/checks.rs
@@ -57,21 +57,18 @@ pub fn run(lines: &[LineEntry], skip_checks: &[&str]) -> Vec<Warning> {
     let mut warnings: Vec<Warning> = Vec::new();
 
     for line in lines {
-        let is_comment = line.is_comment();
-        if is_comment {
-            if let Some(comment) = comment::parse(&line.raw_string) {
-                if comment.is_disabled() {
-                    // Disable checks from a comment using the dotenv-linter:off flag
-                    disabled_checks.extend(comment.checks);
-                } else {
-                    // Enable checks if the comment has the dotenv-linter:on flag
-                    disabled_checks.retain(|&s| !comment.checks.contains(&s));
-                }
+        if let Some(comment) = line.get_comment() {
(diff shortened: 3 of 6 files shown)


mgrachev:
@dotenv-linter/core What do you think about these changes in the `UnorderedKey` check?

src/common/line_entry.rs
```diff
@@ -49,6 +51,15 @@ impl LineEntry {
     pub fn is_last_line(&self) -> bool {
         self.file.total_lines == self.number
     }
+
+    // Maybe we should add the comment field to the LineEntry struct (but this requires some
+    // refactoring of the line entries creation)
+    pub fn get_comment(&self) -> Option<Comment> {
```
mgrachev, line 57:
```suggestion
    pub fn get_control_comment(&self) -> Option<Comment> {
```

README.md
```diff
@@ -449,6 +449,19 @@ FOO=BAR
 BAR=FOO
 ```
 
+Control comments also split lines (this is done to make the linter logic more predictable):
```
mgrachev, line 452: Please add this information to this [page](https://github.com/dotenv-linter/dotenv-linter/blob/master/docs/checks/unordered_key.md) too.

src/checks.rs
```diff
@@ -57,21 +57,18 @@ pub fn run(lines: &[LineEntry], skip_checks: &[&str]) -> Vec<Warning> {
...
-        if is_comment {
-            if let Some(comment) = comment::parse(&line.raw_string) {
-                if comment.is_disabled() {
-                    // Disable checks from a comment using the dotenv-linter:off flag
-                    disabled_checks.extend(comment.checks);
-                } else {
-                    // Enable checks if the comment has the dotenv-linter:on flag
-                    disabled_checks.retain(|&s| !comment.checks.contains(&s));
-                }
+        if let Some(comment) = line.get_comment() {
```
mgrachev, line 60:
```suggestion
        if let Some(comment) = line.get_control_comment() {
```

src/checks/unordered_key.rs
```diff
@@ -27,8 +27,8 @@ impl Default for UnorderedKeyChecker<'_> {
 
 impl Check for UnorderedKeyChecker<'_> {
     fn run(&mut self, line: &LineEntry) -> Option<Warning> {
-        // Support of grouping variables through blank lines
-        if line.is_empty() {
+        // Support of grouping variables through blank lines and control comments
+        if line.is_empty() || line.get_comment().is_some() {
```
mgrachev, line 31:
```suggestion
        if line.is_empty() || line.is_comment() {
```
This check should also support usual comments.

src/checks/unordered_key.rs
```diff
@@ -188,4 +188,15 @@ mod tests {
 
         run_unordered_tests(asserts);
     }
+
+    #[test]
+    fn one_unordered_key_with_control_comment_test() {
+        let asserts = vec![
+            (line_entry(1, 3, "FOO=BAR"), None),
+            (line_entry(2, 3, "# dotenv-linter:off LowercaseKey"), None),
```
mgrachev, line 196: Please also add a test for usual comments.

mgrachev (commented):
@evgeniy-r Thank you for your contribution! 👍

I think we should also support the following case and don't show any warnings:
```env
BB=1
# Simple comment
AA=B
X=X
```

src/checks/unordered_key.rs
evgeniy-r: About usual comments.

In the `UnorderedKey` fixer we now handle a significant line (with key) with all previous comments as a whole, when we sort the keys.

E. g.:
```
B=C
# Comment about A variable
A=B
```
will be fixed to:
```
# Comment about A variable
A=B
B=C
```

I will have to remove this behavior, if we use usual comments as group separators.

Is this OK?

mgrachev: I think yes.

src/checks/unordered_key.rs
evgeniy-r: Done

src/checks.rs
evgeniy-r: Done

src/common/line_entry.rs
evgeniy-r: Done

README.md
evgeniy-r: Done

src/checks/unordered_key.rs
DDtKey: @mgrachev I think this is ambiguous.
This question was raised earlier, as far as I know.

It seems to me not entirely correct to use simple comments as a separator.

For example, if we use comments as separator 
```
DIR=opt
# Comment to FOO line: FOO is BAR
FOO=BAR
BAR=FOO

```
this option will be fixed as:
```
DIR=opt
# Comment to FOO line: FOO is BAR
BAR=FOO
FOO=BAR

```

That is, a comment intended for a specific line will lose its meaning.

Yes, this also can be
ambiguous if comment really intended for group of lines, but user can separate group comment via blank line.

But if we will use usual comments as a separator, we will have trouble with comments to specific lines

README.md
mgrachev, line 452:
```suggestion
Control comments also split lines (this is done to make the linter logic more predictable, will be available in [v2.2.0](https://github.com/dotenv-linter/dotenv-linter/issues/238)):
```

docs/checks/unordered_key.md
```diff
@@ -24,3 +24,16 @@ FOO=BAR
 
 BAR=FOO
 ```
+
+Control comments also split lines (this is done to make the linter logic more predictable):
```
mgrachev, line 28:
```suggestion
Control comments also split lines (this is done to make the linter logic more predictable, will be available in [v2.2.0](https://github.com/dotenv-linter/dotenv-linter/issues/238)):
```

src/checks/unordered_key.rs
mgrachev: @DDtKey I think you are right 👍 I missed this case 🤔

mgrachev: @DDtKey What do you think about splitting lines into groups using control comments?

DDtKey: > @DDtKey What do you think about splitting lines into groups using control comments?

I think this is more of a necessity, we cannot allow reordering of lines beyond the boundaries of comments.

In fact, using the control comments, the user creates the group himself, with additional conditions to processing (ignore specific checks)

mgrachev (approved):

DDtKey (approved):
Thanks for your contribution! :rocket:

PR #289, opened 2020-09-28
Remove code related to Fixed/Unfixed

mgrachev:
#### ✔ Checklist:

@evgeniy-r Please take a look 👀

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -20,31 +20,22 @@ fn main() -> Result<(), Box<dyn Error>> {
         process::exit(0);
     }
 
-    let is_fix = args.is_present("fix");
+    let total = warnings.len();
+    let is_not_quiet = !args.is_present("quiet");
 
-    if is_fix {
-        let (fixed, unfixed): (Vec<_>, Vec<_>) = warnings.iter().partition(|w| w.is_fixed);
-
-        if !fixed.is_empty() {
-            println!("Fixed warnings:");
-            for w in fixed {
-                println!("{}", w);
-            }
+    if args.is_present("fix") {
+        if is_not_quiet {
+            warnings.iter().for_each(|w| println!("{}", w));
+            println!();
         }
 
-        if !unfixed.is_empty() {
-            println!("\nUnfixed warnings:");
-            for w in unfixed {
-                println!("{}", w);
-            }
-        } else {
-            process::exit(0);
-        }
+        println!("All warnings are fixed. Total: {}", total);
+        process::exit(0);
     } else {
         warnings.iter().for_each(|w| println!("{}", w));
 
-        if !args.is_present("quiet") {
-            print_total(warnings.len());
+        if is_not_quiet {
+            print_total(total);
         }
     }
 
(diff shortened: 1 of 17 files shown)


src/main.rs
```diff
@@ -20,32 +20,13 @@ fn main() -> Result<(), Box<dyn Error>> {
         process::exit(0);
     }
 
-    let is_fix = args.is_present("fix");
+    warnings.iter().for_each(|w| println!("{}", w));
```
DDtKey, line 23: I think we shouldn't print warnings for `quiet` mode (or maybe even by default?), because we fix all the `Warning`s.

This seems to be the standard linter behavior, and we also have a flag to configure additiinal output :thinking:
What do you think?

src/main.rs
```diff
@@ -20,32 +20,13 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-        warnings.iter().for_each(|w| println!("{}", w));
-
-        if !args.is_present("quiet") {
-            print_total(warnings.len());
-        }
+    if args.is_present("fix") {
+        println!("\nAll warnings are fixed");
+        process::exit(0);
+    } else if !args.is_present("quiet") {
+        print_total(warnings.len());
```
evgeniy-r, line 29: Should we print the total for fixed warnings too?

src/main.rs
evgeniy-r: I'd like do not print them for quiet mode.

src/main.rs
mstruebing: I would say yes, it's always nice to see some stats when running a tool.

src/main.rs
mgrachev: How do you like these options?

Not in quiet mode:
```bash
$ dotenv-linter -f
Original file was backed up to: ".env_1601369083"

.env:3 LowercaseKey: The a0 key should be in uppercase
.env:4 LowercaseKey: The a2 key should be in uppercase

All warnings are fixed. Total: 2
```

In quiet mode:
```bash
$ dotenv-linter -f --quiet
Original file was backed up to: ".env_1601369083"

All warnings are fixed. Total: 2
```

mgrachev: > I think we shouldn't print warnings for `quiet` mode (or maybe even by default?), because we fix all the `Warning`s.

I think the user should know what warnings were found and fixed without running `dotenv-linter` in check mode.

evgeniy-r: "Not in quiet mode" is looking good.

But maybe don't print anything in quiet mode?

src/main.rs
mgrachev: I like that idea! Thank you! 🙏

src/main.rs
mgrachev: > But maybe don't print anything in quiet mode?

The minimal information always should be in quiet mode.

evgeniy-r: Then "In quiet mode" is OK too.

DDtKey: > How do you like these options?


I have a little doubt about the output of backup files (in the case when there are many files, this is not so "`quiet`")
But yes, the user should know about the created backups (maybe they should just be placed in a special folder?)

So, I like the proposed options, thank you!

mgrachev: > I have a little doubt about the output of backup files (in the case when there are many files, this is not so "`quiet`")
> But yes, the user should know about the created backups (maybe they should just be placed in a special folder?)

If you have an idea how we can improve the output of backup files, please open an issue and we will discuss it.

DDtKey (dismissed):
Thank you for your contribution! :+1:

PR #311, opened 2020-10-06
Shows files that were linted

### LINKED ISSUE #268: Display file names being scanned
Since the program doesn't output anything when files don't have warnings, as a user, I would like to see which files have been scanned, especially in the case of a deeply nested recursive situation.

```
Linting foo.env
Linting nested/bar.env
Linting nested/dir/test.env
```

I'm happy to take this on if owners feel like this would be a useful addition to the project.

Anthuang:
This PR adds functionality to show the files that were linted as part of a dotenv-linter run. It does so by introducing a new struct `Warnings` which is basically a wrapper around `Vec<Warning>` to also include information about the paths related to the warnings.

This PR also fixes tests as a result of the added functionality, and some refactoring.

Below shows the output from running dotenv-linter with different args:
```
$ cargo run
Checking ".env"
.env:1 LowercaseKey: The hi key should be in uppercase
.env:3 ExtraBlankLine: Extra blank line detected
.env:4 LowercaseKey: The hi3 key should be in uppercase

Checking ".env_2/.env"
.env_2/.env:2 ExtraBlankLine: Extra blank line detected
.env_2/.env:3 LowercaseKey: The hi key should be in uppercase
.env_2/.env:4 DuplicatedKey: The hi key is duplicated
.env_2/.env:4 LowercaseKey: The hi key should be in uppercase

Found 7 problems
```

```
$ cargo run -- -q
.env:1 LowercaseKey: The hi key should be in uppercase
.env:3 ExtraBlankLine: Extra blank line detected
.env:4 LowercaseKey: The hi3 key should be in uppercase
.env_2/.env:2 ExtraBlankLine: Extra blank line detected
.env_2/.env:3 LowercaseKey: The hi key should be in uppercase
.env_2/.env:4 DuplicatedKey: The hi key is duplicated
.env_2/.env:4 LowercaseKey: The hi key should be in uppercase
```

```
$ cargo run -- -f
Checking ".env"
Original file was backed up to: ".env_1601943961"

.env:1 LowercaseKey: The hi key should be in uppercase
.env:3 ExtraBlankLine: Extra blank line detected
.env:4 LowercaseKey: The hi3 key should be in uppercase

Checking ".env_2/.env"
Original file was backed up to: ".env_2/.env_1601943961"

.env_2/.env:2 ExtraBlankLine: Extra blank line detected
.env_2/.env:3 LowercaseKey: The hi key should be in uppercase
.env_2/.env:4 DuplicatedKey: The hi key is duplicated
.env_2/.env:4 LowercaseKey: The hi key should be in uppercase

All warnings are fixed. Total: 7
```

```
$ cargo run -- -f -q
Original file was backed up to: ".env_1601943979"
Original file was backed up to: ".env_2/.env_1601943979"

All warnings are fixed. Total: 7
```

```
$ cargo run -- -f --no-backup
Checking ".env"
.env:1 LowercaseKey: The hi key should be in uppercase
.env:3 ExtraBlankLine: Extra blank line detected
.env:4 LowercaseKey: The hi3 key should be in uppercase

Checking ".env_2/.env"
.env_2/.env:2 ExtraBlankLine: Extra blank line detected
.env_2/.env:3 LowercaseKey: The hi key should be in uppercase
.env_2/.env:4 DuplicatedKey: The hi key is duplicated
.env_2/.env:4 LowercaseKey: The hi key should be in uppercase

All warnings are fixed. Total: 7
```

```
$ cargo run -- -f -q --no-backup

All warnings are fixed. Total: 7
```

#### ✔ Checklist:

Closes #268


Anthuang:
I think the bug is a bug from cargo itself?

mgrachev:
@Anthuang Thank you for your contribution! 🔥

> I think the error is a bug from cargo itself?

I think not 🤔
```
failures:

---- common::warnings::tests::warnings_fmt_test stdout ----
---- common::warnings::tests::warnings_fmt_test stderr ----
thread 'main' panicked at 'Could not create FileEntry', src/common/warnings.rs:71:21
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Anthuang:
> @Anthuang Thank you for your contribution! 🔥
> 
> > I think the error is a bug from cargo itself?
> 
> I think not 🤔
> 
> ```
> failures:
> 
> ---- common::warnings::tests::warnings_fmt_test stdout ----
> ---- common::warnings::tests::warnings_fmt_test stderr ----
> thread 'main' panicked at 'Could not create FileEntry', src/common/warnings.rs:71:21
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```

I made the mistake of not scrolling up haha

src/common/warnings.rs
```diff
@@ -0,0 +1,76 @@
...
+    #[test]
+    fn warnings_fmt_test() {
+        let line = line_entry(1, 1, "FOO=BAR");
+        let warning = Warning::new(
+            line,
+            "DuplicatedKey",
+            String::from("The FOO key is duplicated"),
+        );
+        let fe = blank_file_entry(1);
+        let warnings = Warnings::new(fe, None, vec![warning]);
```
mgrachev, line 69:
```suggestion
            line.clone(),
            "DuplicatedKey",
            String::from("The FOO key is duplicated"),
        );
        let warnings = Warnings::new(line.file, None, vec![warning]);
```

src/common.rs
```diff
@@ -20,6 +22,15 @@ pub(crate) mod tests {
     use super::*;
     use std::path::PathBuf;
 
+    #[allow(dead_code)]
+    pub fn blank_file_entry(total_lines: usize) -> FileEntry {
```
mgrachev, line 26: I don't think that we need this function. See the comment above.

src/common/warnings.rs
```diff
@@ -0,0 +1,76 @@
+use crate::common::{FileEntry, Warning};
+use std::ffi::OsString;
+use std::fmt;
+
+/// Wraps warnings to provide more information when printing.
+pub struct Warnings {
```
mgrachev, line 6: I think this struct should be named `Output`. What do you think?
/cc @dotenv-linter/core

src/common/warnings.rs
```diff
@@ -0,0 +1,76 @@
...
+            println!("Original file was backed up to: {:?}", p);
+        }
+    }
+}
+
+impl fmt::Display for Warnings {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Checking {}", self.path)?;
+        if let Some(p) = &self.backup_path {
+            writeln!(f, "\nOriginal file was backed up to: {:?}", p)?;
```
mgrachev, line 43: I see the duplicated line https://github.com/dotenv-linter/dotenv-linter/pull/311/files#diff-cefc832d3c36ffe8734f777f9d768fa7R34. Why is this needed?

src/main.rs
```diff
@@ -20,25 +20,36 @@ fn main() -> Result<(), Box<dyn Error>> {
         process::exit(0);
     }
 
-    let total = warnings.len();
+    let mut total = 0;
+    warnings.iter().for_each(|w| total += w.warnings.len());
```
mgrachev, line 24: I think you can move the counting of totals to the `Output` struct 🤔

src/main.rs
```diff
@@ -20,25 +20,36 @@ fn main() -> Result<(), Box<dyn Error>> {
...
     let is_not_quiet = !args.is_present("quiet");
 
     if args.is_present("fix") {
+        // Don't display anything when there are no warnings
+        if total == 0 {
+            process::exit(0);
```
mgrachev, line 30: I think you can do it earlier before `args.is_present("fix")`.

tests/common/output.rs
```diff
@@ -1,13 +1,56 @@
-const FIX_SENTENCE: &str = "All warnings are fixed. Total:";
+enum Mode {
+    Fix,
+    Check,
+}
+
+/// Builds test output for validation.
+fn build_output(blocks: &[(&str, &[&str])], mode: Mode) -> String {
+    let mut output = "".to_string();
```
mgrachev, line 8:
```suggestion
    let mut output = String::new();
```

mgrachev (commented):
@Anthuang Thank you for your contribution! 🔥

I have left some comments. Please look at them 👀

src/common/warnings.rs
Anthuang: `Output` does sound better, but I think it fits more if that were the name for `Vec<Warnings>`. Each `Warnings` is meant to be a file and the errors it contains.

src/main.rs
Anthuang: This would be counting across `Output`s, since `run` gives a `Vec` of `Output`s.

src/main.rs
Anthuang: I realized I had this in the wrong spot. To support the case where we still output "Checking FILE" even when there are no warnings at all, we'd have to do this check later instead of earlier.

In the case of no warnings, should we output text to indicate that? I.e. instead of "All warnings are fixed. Total:" we say "No warnings found".

src/common.rs
Anthuang: Will remove!

src/common/warnings.rs
Anthuang: Ah thanks!

tests/common/output.rs
Anthuang: 👍

src/common/warnings.rs
Anthuang: In order to support printing backup files without all the warnings. See below:
```
$ cargo run -- -f -q
Original file was backed up to: ".env_1601943979"
Original file was backed up to: ".env_2/.env_1601943979"

All warnings are fixed. Total: 7
```

Is there a better way of sharing the code to print the backups?

Anthuang:
Thanks for the review!!

src/main.rs
mgrachev: I like that idea 🙂

src/common/warnings.rs
mgrachev: I need to think about it 🤔

src/main.rs
```diff
@@ -14,42 +14,56 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+    }
     process::exit(1);
 }
 
-fn print_total(total: usize) {
+fn print_fix_total(total: usize) {
+    if total != 0 {
+        println!("\nAll warnings are fixed. Total: {}", total);
+    } else {
+        println!("\nNo warnings found");
```
mgrachev, line 55: You should print this message in the `print_check_total` function too.
Right now it's printing `Found 0 problems`.
And please add tests for these cases 🙏

src/main.rs
```diff
@@ -14,42 +14,56 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+fn print_check_total(total: usize) {
     let mut problems = String::from("problem");
 
-    if total > 1 {
+    if total != 1 {
         problems += "s";
     }
 
-    println!("\n{}", format!("Found {} {}", total, problems));
+    println!("{}", format!("\nFound {} {}", total, problems));
```
mgrachev, line 66: I got the following result:
```
Checking .env
.env:4 DuplicatedKey: The A2 key is duplicated
.env:4 LeadingCharacter: Invalid leading character detected


Found 2 problems
```

You can see 2 blank lines above the `Found 2 problems` message. Can you fix it?
And it would be great if you wrote a test for this case.

tests/common/output.rs
```diff
@@ -1,13 +1,60 @@
...
+    Check,
+}
+
+/// Builds test output for validation.
+fn build_output(blocks: &[(&str, &[&str])], mode: Mode) -> String {
+    let mut output = String::new();
+    let mut count = 0;
+
+    for b in blocks {
+        output.push_str(&format!("Checking {}\n", b.0));
```
mgrachev, line 12: Please add a comment what it is`b.0`.

tests/common/output.rs
```diff
@@ -1,13 +1,60 @@
...
+        if !b.1.is_empty() {
```
mgrachev, line 13: Please add a comment what it is`b.1`.

tests/flags/quiet.rs
```diff
@@ -20,7 +20,7 @@ fn fix_output_in_quiet_mode() {
     let _ = test_dir.create_testfile(".env", "abc=DEF\n\nF=BAR\nB=bbb\n");
 
     let args = &["--quiet"];
-    let expected_output = format!("All warnings are fixed. Total: {}\n", 2);
+    let expected_output = format!("\nAll warnings are fixed. Total: {}\n", 2);
```
mgrachev, line 23: Is that necessary?

I got the result with 2 blank lines before this message:
```
Checking .env
Original file was backed up to: ".env_1602361743"

.env:4 DuplicatedKey: The A2 key is duplicated
.env:4 LeadingCharacter: Invalid leading character detected


All warnings are fixed. Total: 2
```

Anthuang: Ah, good catch.

src/main.rs
```diff
@@ -14,42 +14,56 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+    let outputs = dotenv_linter::run(&args, &current_dir)?;
 
-    if warnings.is_empty() {
+    if outputs.is_empty() {
         process::exit(0);
     }
 
-    let total = warnings.len();
+    let mut total = 0;
+    outputs.iter().for_each(|w| total += w.warnings.len());
```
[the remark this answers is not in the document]
Anthuang: Cool!

src/common/output.rs
```diff
@@ -0,0 +1,75 @@
...
+        }
+    }
+}
+
+impl fmt::Display for Output {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "Checking {}", self.path)?;
+        if let Some(p) = &self.backup_path {
+            writeln!(f, "\nOriginal file was backed up to: {:?}", p)?;
+        }
```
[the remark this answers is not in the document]
Anthuang: Looks good to me!

src/common/output.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    #[test]
+    fn output_fmt_test() {
+        let line = line_entry(1, 1, "FOO=BAR");
+        let warning = Warning::new(
+            line.clone(),
+            "DuplicatedKey",
+            String::from("The FOO key is duplicated"),
+        );
+        let output = Output::new(line.file, None, vec![warning]);
```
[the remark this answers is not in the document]
Anthuang: Is there a good way to test stdout output in unit tests? Or is it better to do this in the tests directory? It seems like the current tests always pass "--no-backup".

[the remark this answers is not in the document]
Anthuang: I was able to get the backup files to work in tests, but it is hard to validate the output since the backup filenames depend on the system time. I could add an option to `dotenv-linter` that sets a specific suffix to the backed up files, which would also let me test this.

Instead of:
```rust
pub fn backup_file(fe: &FileEntry) -> Result<PathBuf, Box<dyn Error>>
```
I could create another function:
```rust
pub fn backup_file_with_suffix(fe: &FileEntry, suffix: String) -> Result<PathBuf, Box<dyn Error>>
```
If no suffix is specified then `backup_file` will just call `backup_file_with_suffix` with the timestamp as the suffix. If the user specifies a suffix, let's say through an argument like `--backup-suffix`, then it will call `backup_file_with_suffix` directly.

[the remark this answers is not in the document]
mgrachev: > I was able to get the backup files to work in tests, but it is hard to validate the output since the backup filenames depend on the system time. I could add an option to `dotenv-linter` that sets a specific suffix to the backed up files, which would also let me test this.
> 
> Instead of:
> 
> ```rust
> pub fn backup_file(fe: &FileEntry) -> Result<PathBuf, Box<dyn Error>>
> ```
> 
> I could create another function:
> 
> ```rust
> pub fn backup_file_with_suffix(fe: &FileEntry, suffix: String) -> Result<PathBuf, Box<dyn Error>>
> ```
> 
> If no suffix is specified then `backup_file` will just call `backup_file_with_suffix` with the timestamp as the suffix. If the user specifies a suffix, let's say through an argument like `--backup-suffix`, then it will call `backup_file_with_suffix` directly.

I wouldn't like to add an additional complexity to this method.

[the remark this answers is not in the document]
mgrachev: > Is there a good way to test stdout output in unit tests? Or is it better to do this in the tests directory? It seems like the current tests always pass "--no-backup".

I would like to have the unit test like this:
```rust
    #[test]
    fn fix_output_fmt_test() {
        let line = line_entry(1, 1, "FOO=BAR");
        let warning = Warning::new(
            line.clone(),
            "DuplicatedKey",
            String::from("The FOO key is duplicated"),
        );

        let backup_path = OsString::from(".env_1234");
        let output = Output::new(line.file, Some(backup_path), vec![warning]);

        assert_eq!(
            "Fixing .env\nOriginal file was backed up to: \".env_1234\"\n\n.env:1 DuplicatedKey: The FOO key is duplicated\n",
            format!("{}", output)
        );
    }
```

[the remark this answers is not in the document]
mgrachev: > Or is it better to do this in the tests directory? It seems like the current tests always pass "--no-backup".

Also, I would like to have several integration tests to check different outputs options, because I have already found several cases with extra blank lines in outputs. In these integration tests please don't use the `build_output` function.

> I was able to get the backup files to work in tests, but it is hard to validate the output since the backup filenames depend on the system time.

I think it can help you https://github.com/dotenv-linter/dotenv-linter/blob/master/tests/flags/backup.rs

Anthuang:
I messed up the git history a bit when rebasing, hopefully this fixes it

src/common/output.rs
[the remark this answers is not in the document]
Anthuang: Thank you for the review! I have added the unit test, and tweaked the `print_backup` function to work with it. I've also added the integration tests.

For integration testing the backup, I'm not sure how to validate the stdout easily since the backup file name is dependent on the timestamp.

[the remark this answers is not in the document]
mgrachev: You can do something like this:
```rust
let backup_file = fs::read_dir(&testdir.as_str())
    .expect("read dir")
    .filter_map(|e| e.ok())
    .filter(|e| e.path().as_os_str() != testfile.as_str())
    .last()
    .expect("get backup file");

let backup_filename = backup_file.file_name();
```

src/common.rs
```diff
@@ -1,12 +1,20 @@
...
+mod output;
 mod warning;
 
 pub use file_entry::FileEntry;
 pub use line_entry::LineEntry;
+pub use output::Output;
 pub use warning::Warning;
 
+/// Mode in which the program is run.
+pub enum Mode {
```
mgrachev, line 13: Please move this `enum` to `common/output.rs`.

tests/checks/correct_output.rs
```diff
@@ -0,0 +1,146 @@
+//! Tests that output from checks are correct. Mainly needed to ensure that
```
mgrachev, line 1: I would move this file to `tests/output/check.rs`.

src/lib.rs
```diff
@@ -51,25 +51,39 @@ pub fn run(args: &clap::ArgMatches, current_dir: &PathBuf) -> Result<Vec<Warning
...
 
-    Ok(warnings)
+    Ok(outputs)
+}
+
+/// Prints outputs with correct newlines.
+///
+/// Prints a newline after an output if it is not the last output or, for the
+/// last output, it has no warnings.
+pub fn print_outputs(outputs: Vec<Output>) {
```
mgrachev, line 80: To tell the truth, I see that you should do some refactoring.
I sugger to create the `Output` struct, which will store all output and has this method.
In addition, I suggest separating the current `Output` structure into `Check` and `Fix`.
It will allow to get rid of `if/else` conditions in many files.
What do you think?

src/main.rs
```diff
@@ -14,42 +14,59 @@ fn main() -> Result<(), Box<dyn Error>> {
...
+    }
 
-        if is_not_quiet {
-            print_total(total);
-        }
+    if is_not_quiet {
+        dotenv_linter::print_outputs(outputs);
+        print_check_total(total);
+    } else {
+        outputs.iter().for_each(|w| w.print_warnings());
```
mgrachev, line 40: I would like to have the only one method instead of `if/else` condition. For example:
```rust
let is_quiet = args.is_present("quiet");
output.print_check_output(total, is_quiet);
```

tests/common/output.rs
```diff
@@ -1,13 +1,83 @@
-const FIX_SENTENCE: &str = "All warnings are fixed. Total:";
+enum Mode {
```
mgrachev, line 1: You don't need to create this `enum` - please use already defined.

tests/common/output.rs
```diff
@@ -1,13 +1,83 @@
...
+///
+/// * `blocks` - A slice of tuples, each containing a file path and a slice of
+/// warnings for that file
+/// * `mode` - Mode in which the program is run.
+fn build_output(blocks: &[(&str, &[&str])], mode: Mode) -> String {
+    let mut output = String::new();
+    let mut count = 0;
+
+    if blocks.is_empty() {
+        return String::new();
```
mgrachev, line 18:
```suggestion
        return output;
```

tests/fixes/correct_output.rs
```diff
@@ -0,0 +1,171 @@
+//! Tests that output from fixes are correct. Mainly needed to ensure that
```
mgrachev, line 1: I would move this file to `tests/output/fix.rs`.

tests/fixes/correct_output.rs
```diff
@@ -0,0 +1,171 @@
...
+    test_dir.test_command_fix_success_with_args(expected_output, args);
+    test_dir.close();
+}
+
+#[test]
+fn quiet_no_warnings() {
+    let test_dir = TestDir::new();
+    test_dir.create_testfile(".env", "ABC=DEF\nB=bbb\nF=BAR\n");
+
+    let args = &["--quiet"];
```
mgrachev, line 162: I don't see tests with the backup feature 🤔

tests/common/output.rs
Anthuang: I did this to avoid having to make the `common` module (as well as `common/output`) public in the crate. Is it okay to make them public?

tests/fixes/correct_output.rs
Anthuang: Will add them!

tests/common/output.rs
mgrachev: Let's leave it as it is.

src/lib.rs
mgrachev: This PR is already pretty big. I think we can do it later.

src/main.rs
mgrachev: We will do it next time.

mgrachev:
@Anthuang Please fix conflicts.

Anthuang:
> @Anthuang Please fix conflicts.

Done!

mgrachev:
@Anthuang Thank you! I'm going to merge this PR after v2.2.1 release.

src/lib.rs
```diff
@@ -51,25 +51,43 @@ pub fn run(args: &clap::ArgMatches, current_dir: &PathBuf) -> Result<Vec<Warning
...
-                println!("Original file was backed up to: {:?}\n", backup_file);
+                backup_file = Some(fs_utils::backup_file(&fe)?.into_os_string());
             }
 
             // write corrected file
             fs_utils::write_file(&fe.path, lines)?;
         }
 
-        warnings.extend(result);
+        let mode = if is_fix {
```
DDtKey, line 69: I think we can do this once before the loop, when we read `--fix` argument.

DDtKey (commented):
Thank you for your contribution! :+1: 

To be honest, this is not exactly what I imagined after #268 :slightly_frowning_face: 
It seems to me that it was about interactive mode.
When we see the progress of checking files on the fly (many CLI utilities do just that)

Perhaps the `dotenv-linter` works very quickly and even with a large number of files it is not necessary, but in this case, it generally does not make much sense, except for displaying the files that have been checked :thinking: 
I am ready to listen to any opinions on this matter :eyes: 

What do you say about it @Anthuang  @dotenv-linter/core  @gillespiecd ?

Either way, moving some of the output logic into a separate module is a great idea!

mgrachev:
> It seems to me that it was about interactive mode.
> When we see the progress of checking files on the fly (many CLI utilities do just that)

You are absolutely right. This is just the first iteration.
We can add the progress of checking files on the fly later.

> Perhaps the `dotenv-linter` works very quickly and even with a large number of files it is not necessary, but in this case, it generally does not make much sense, except for displaying the files that have been checked(which is also important) 🤔

Right now `dotenv-linter` does not output anything if it did not find problems in the `.env` files.
In such situation, it is not obvious to the user whether `dotenv-linter` has checked all the files or not.
So, this PR fixes that case.

DDtKey:
> You are absolutely right. This is just the first iteration.
> We can add the progress of checking files on the fly later.

I completely agree with the division of the task into iterations! :+1: 

> Right now `dotenv-linter` does not output anything if it did not find problems in the `.env` files.
> In such situation, it is not obvious to the user whether `dotenv-linter` has checked all the files or not.
> So, this PR fixes that case.

Then it might be more correct at the moment to write `Checked {}` instead of `Checking {}`, since the output occurs after all the checks are completed?
But in general, it is not necessary that in the future the structure of the output does not change much with the appearance of the output "on the fly" :slightly_smiling_face:

mgrachev:
> But in general, it is not necessary that in the future the structure of the output does not change much with the appearance of the output "on the fly" 🙂

@DDtKey Maybe you would like to implement on-the-fly output? 😉

DDtKey:
> > But in general, it is not necessary that in the future the structure of the output does not change much with the appearance of the output "on the fly" 🙂
> 
> 
> 
> @DDtKey Maybe you would like to implement on-the-fly output? 😉

Why not, I'll gladly take it when we merge to master 🙂

src/lib.rs
mgrachev: @Anthuang Can you make these changes?

mgrachev:
@Anthuang Please fix linter's warnings.

mgrachev:
@Anthuang Thanks a lot! 🔥

Anthuang:
Thanks for the reviews!

Issue #335, opened 2020-10-26
Make an output on-the-fly


DDtKey:
Make output of `checking`/`fixing` files and results on-the-fly(during `dotenv-linter` work), instead of output at the end of the program.

It is planned to keep the structure of the current output(just add interactivity)

PR #336, opened 2020-10-28
feat: make an output on-the-fly(#335)

DDtKey:
I thought about several options(_up to the reactive approach, which is overkill in the linter_), but it seemed to me that such a solution would be rather transparent and linear in the code.
By and large, the program output itself is completely taken out in a separate module.

Also, the structure of the output was completely preserved and the **integration tests remained relevant!**

And I noticed that we didn't update the documentation after adding the `checking`/`fixing` files output.
I will update this here and in a related project (site), but first I wanted to know your opinion on the implementation.
**UPD**: [PR to site](https://github.com/dotenv-linter/dotenv-linter.github.io/pull/11)

Maybe there are ideas or suggestions with other approaches? @dotenv-linter/core

#### ✔ Checklist:


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Display linted files when run [#311](https://github.com/dotenv-linter/dotenv-linter/pull/311) ([@Anthuang](https://github.com/anthuang))
 
 ### 🔧 Changed
+- feat: make a output on-the-fly [#336](https://github.com/dotenv-linter/dotenv-linter/pull/336) ([@DDtKey](https://github.com/DDtKey))
```
mgrachev, line 12:
```suggestion
- Make a output on-the-fly [#336](https://github.com/dotenv-linter/dotenv-linter/pull/336) ([@DDtKey](https://github.com/DDtKey))
```

src/common/output.rs
```diff
@@ -13,107 +12,90 @@ pub enum Mode {
 const BACKUP_PREFIX: &str = "Original file was backed up to: ";
 
 /// Wraps warnings to provide more information when printing.
+#[derive(Clone, Copy)]
 pub struct Output {
```
mgrachev, line 16: I think we should split this module to several: `Check` & `Fix`.
In the future we will add more commands like `compare` and for these commands we will have to create separate outputs.

mgrachev (commented):
@DDtKey You have done excellect work! 🔥

src/common/output.rs
DDtKey: I like this idea 👍
I'll come back to this soon.

DDtKey: > I think we should split this module to several: `Check` & `Fix`.

In fact, the method of separation was not so obvious.
The calling code for `Fix` and `Check` is strongly interconnected, which I think cannot be said about `Compare`.
So I had to add some conditions/logic to `dotenv_linter::run`, but I tried to make it the most obvious and concise way.

In general, I considered various options: traits, general output configuration structure, etc. But it looks a bit cumbersome for simple output functions and somehow framed `Fix` and `Check` into something combined, complicating understanding in the case of adding a `compare`.

Thus, we have split modules with their own methods. Please take a look at this 👀 
I will be glad to hear any suggestions and other options 🙂

src/lib.rs
```diff
@@ -55,39 +42,68 @@ pub fn run(args: &clap::ArgMatches, current_dir: &PathBuf) -> Result<Vec<Output>
...
             }
 
             // write corrected file
             fs_utils::write_file(&fe.path, lines)?;
         }
 
-        outputs.push(Output::new(fe, backup_file, result, mode));
+        // This shouldn't be printed to Fox when combined with quiet mode
+        if !(is_fix && is_quiet_mode) {
+            let is_last_file = i == file_paths.len() - 1;
```
mgrachev, line 69: Can we move this code to the `check_output.print_warnings` method?

src/main.rs
```diff
@@ -15,74 +14,23 @@ fn main() -> Result<(), Box<dyn Error>> {
...
-        process::exit(0);
-    }
 
-    if is_not_quiet {
-        dotenv_linter::print_outputs(outputs);
-        print_check_total(total);
-    } else {
-        outputs.iter().for_each(|w| w.print_warnings());
-    }
+    let warnings = dotenv_linter::run(&args, &current_dir)?;
```
mgrachev, line 24: Maybe it's better to return the number of warnings instead of an array of warnings?

src/lib.rs
```diff
@@ -55,39 +42,68 @@ pub fn run(args: &clap::ArgMatches, current_dir: &PathBuf) -> Result<Vec<Output>
...
+                let backup_file = fs_utils::backup_file(&fe)?.into_os_string();
+                fix_output.print_backup(&backup_file);
             }
 
             // write corrected file
             fs_utils::write_file(&fe.path, lines)?;
         }
 
-        outputs.push(Output::new(fe, backup_file, result, mode));
+        // This shouldn't be printed to Fox when combined with quiet mode
```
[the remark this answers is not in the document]
DDtKey: So planned 😹

src/lib.rs
DDtKey: > Can we move this code to the `check_output.print_warnings` method?

We could make the same method in `fix_output`, but it is not entirely correct to add it inside the `check_output` module.
Thus, I just decided to reduce the amount of duplicate code, perhaps it is better to duplicate than to complicate the logic and it will be better 🤔

DDtKey: Ie we can make something like this:
```
        if is_fix {
            fix_output.print_warnings(&result, is_last_file);
        } else {
            check_output.print_warnings(&result, is_last_file)
        }
```

or passing a parameter to the enabled commit, which will confuse the logic a bit (but will reduce conditions):
```
     check_output.print_warnings(&result, fix_enabled, is_last_file);
```

@mgrachev what do you think? 🤔

mgrachev: I meant this line:
```rust
let is_last_file = i == file_paths.len() - 1;
```
Can we define the last line in the `check_output.print_warnings` method?
If not, let's leave it as it was 🙂

DDtKey: > I meant this line:
> 
> ```rust
> let is_last_file = i == file_paths.len() - 1;
> ```
> 
> Can we define the last line in the `check_output.print_warnings` method?

Oh, sorry for misunderstanding 🙂 
I made a commit, please take a look at implementation option for this

mgrachev (approved):

src/lib.rs
mgrachev: Excellent implementation 👍

mgrachev:
@DDtKey Thanks a lot! ⚡️

PR #342, opened 2020-11-15
Remade flags to subcommands

### LINKED ISSUE #333: Proposal: Using subcommands intead of flags
### Problems
1. We now have a few flags, which are essentially subcommands: `--fix` & `--show-checks`;
2. Some flags don't work with other. Example: `$ dotenv-linter --no-backup`;
3. Adding new commands like `compare` confuses the code even more.

### Solution
**Breaking changes**: Using subcommands intead of flags.

Checking `.env` files will by default have no subcommand:

```bash
$ dotenv-linter
```

Fixing `.env` files will have the `fix` subcommand:

```bash
$ dotenv-linter fix
```

Showing checks will have the `checks` subcommand:

```bash
$ dotenv-linter checks
```

### Additional changes
Move to the [clap](https://github.com/clap-rs/clap) `v3.0.0-beta.2` - there are a lot of new changes like:
* Generate a CLI simply by defining a struct;
* Auto-generated completion scripts (using `clap_generate`);
* Derive macros;
* Support for building CLIs from YAML;
* Automatic Version from `Cargo.toml`.

What do you think?

mgrachev:
Close #333 

What was done:
- [x] Remade the `--fix/-f` flag to the `fix` subcommand;
- [x] Remade the `--show-checks` flag to the `list` subcommand;
- [x] Fixed a bug with output for multiple files in quiet mode (and wrote a test for that case);
- [x] Added colored help output.

#### ✔ Checklist:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,5 @@
 use crate::common::*;
-
+use std::collections::BTreeMap;
 use std::error::Error;
 use std::path::PathBuf;
 
@@ -10,13 +10,12 @@ mod fs_utils;
 
 pub use checks::available_check_names;
 
-#[allow(clippy::redundant_closure)]
-pub fn run(args: &clap::ArgMatches, current_dir: &PathBuf) -> Result<usize, Box<dyn Error>> {
+pub fn check(args: &clap::ArgMatches, current_dir: &PathBuf) -> Result<usize, Box<dyn Error>> {
     let mut warnings_count = 0;
-    let file_paths: Vec<PathBuf> = get_needed_file_paths(args);
+    let lines_map = get_lines(args, current_dir)?;
 
-    // Nothing to check/fix
-    if file_paths.is_empty() {
+    // Nothing to check
+    if lines_map.is_empty() {
         return Ok(warnings_count);
     }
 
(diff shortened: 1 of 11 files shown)


src/main.rs
```diff
@@ -79,23 +81,66 @@ fn get_args(current_dir: &OsStr) -> clap::ArgMatches {
...
+                .arg(
+                    Arg::with_name("input")
+                        .help("files or paths")
+                        .index(1)
+                        .default_value_os(current_dir)
+                        .required(true)
+                        .multiple(true),
+                )
+                .arg(
+                    Arg::with_name("exclude")
```
DDtKey, line 104: How do you look at make an array with common `check` and `fixed` arguments?
Something like a function `get_common_arguments()` ? 🤔 
As I see it is true for: `input`, `exclude`, `skip`, `recursive`, `no-color`, `quiet`

src/main.rs
```diff
@@ -79,23 +81,66 @@ fn get_args(current_dir: &OsStr) -> clap::ArgMatches {
...
-                .long("no-backup")
-                .help("Prevents .env files from being backed up when modified by -f/--fix"),
-        )
         .arg(
             Arg::with_name("quiet")
                 .short("q")
                 .long("quiet")
                 .help("Doesn't display additional information"),
         )
+        .subcommand(SubCommand::with_name("checks").about("Shows list of available checks"))
```
DDtKey, line 90: It seems to me that this can be a little confusing, especially if we still need a subcommand with the name `check`. Maybe it makes sense to rename to something like: `names` or `list ? Or any another suggestions?
By and large, names are used not only for checks 👀

src/lib.rs
```diff
@@ -42,17 +38,57 @@ pub fn run(args: &clap::ArgMatches, current_dir: &PathBuf) -> Result<usize, Box<
...
-        let mut lines = get_line_entries(&fe, strs);
+        warnings_count += result.len();
+    }
+
+    check_output.print_total(warnings_count);
+
+    Ok(warnings_count)
+}
+
+pub fn fix(args: &clap::ArgMatches, current_dir: &PathBuf) -> Result<(), Box<dyn Error>> {
```
DDtKey, line 58: Now in the fix function there is no output of found errors in the absence of a `quiet` flag.
Is it planned? Tests are failing on my Mac now 🤔

DDtKey (commented):
It looks very transparent now, thanks for the work done! 🚀 
I left a couple of comments, please take a look 🙂 

P.S: **I also see that this is not yet ready**, but I allowed myself to leave comments right now, I hope this can be useful and I do not interfere with the progress of your work 🙏

src/main.rs
mgrachev: I was just about to do it 😃

src/main.rs
mgrachev: I like `list`, thank you! 👍

src/lib.rs
mgrachev: No, it is not planned. I will fix that 😅

src/lib.rs
```diff
@@ -25,60 +24,91 @@ pub fn run(args: &clap::ArgMatches, current_dir: &PathBuf) -> Result<usize, Box<
...
+    output.print_total(warnings_count);
+
+    Ok(())
+}
+
+fn get_lines(
+    args: &clap::ArgMatches,
+    current_dir: &PathBuf,
+) -> Result<BTreeMap<FileEntry, Vec<String>>, Box<dyn Error>> {
+    let mut warnings: BTreeMap<FileEntry, Vec<String>> = BTreeMap::new();
```
DDtKey, line 90: I think it's better to rename this variable 🙂

DDtKey (approved):

PR #347, opened 2020-11-27
Add commitlint

mgrachev:
#### ✔ Checklist:

- [x] This PR has been added to [CHANGELOG.md](https://github.com/dotenv-linter/dotenv-linter/blob/master/CHANGELOG.md) (at the top of the list).

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -10,8 +10,9 @@ Bug fixes and new features should include tests.
 #### ✔ Checklist:
 <!-- Remove items that do not apply. For completed items, change [ ] to [x]. -->
 
+- [ ] Commit messages have been written in [Conventional Commits](https://www.conventionalcommits.org) format;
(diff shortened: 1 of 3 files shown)


mstruebing (approved):
:tada:

What do you think about adding a commithook to already check this while commiting and not having to wait for the pipeline, if not commit with `--no-verify`?

mgrachev:
> What do you think about adding a commithook to already check this while commiting and not having to wait for the pipeline, if not commit with `--no-verify`?

I didn't quite understand you. Can you give an example?

mstruebing:
A git hook is something that hooks into your git lifecycle. So for example, a pre-commit-hook for example will run exactly before you commit. It is a simple bash script which aborts the lifecycle if it exits with a non zero return value.

So you could run tests as a pre-commit hook and don't allow to commit if some tests fail for example.
You could skip the git hooks with `git commit --no-verify`, but you have to want to do it, it will not happen automatically.

In our case we would need a commit-msg hook. There are mostly a bunch of tools for every language available, to make this as easy usable as possible. But in general they are just bash script which are placed in the repository :`.git/hooks/`.

This is the first one I've found: https://github.com/swellaby/rusty-hook

That way, you could see that your commit message is wrong without having to wait for the github action pipelines to finish.

mgrachev:
Thank you! I will think about it.

Issue #384, opened 2021-03-16
Print a message "Nothing to check"


mgrachev:
### Description

Print a message "Nothing to check" in the absence of `.env` files for check.

### Implementation

**Step 1:**

Add a method for [CheckOutput](https://github.com/dotenv-linter/dotenv-linter/blob/master/src/common/output/check.rs#L11) to print the message "Nothing to check" only if `is_quiet_mode == false`.

**Step 2:**

Call the method from the step 1 when [lines_map](https://github.com/dotenv-linter/dotenv-linter/blob/master/src/lib.rs#L20) is empty.

**Step 3:**

Add integration tests [here](https://github.com/dotenv-linter/dotenv-linter/blob/master/tests/output/check.rs).

### Example

There aren't `.env` files in the current directory:

```sh
$ dotenv-linter
Nothing to check
```

jodli:
I will look at this in the evening, if it's okay! :)

mstruebing:
Sure, I assign you to that issue.

PR #392, opened 2021-03-18
Print message if there are no input files (#384)

jodli:
This implements #384 

I had to change the exclude tests as well, as they just excluded all loaded files... :)

#### ✔ Checklist:

### DIFF
--- a/src/common/output/check.rs
+++ b/src/common/output/check.rs
@@ -16,6 +16,17 @@ impl CheckOutput {
         }
     }
 
+    /// Prints information when there is nothing to check and returns a boolean
+    pub fn is_something_to_check(&self) -> bool {
+        if self.files_count == 0 {
+            if !self.is_quiet_mode {
+                println!("Nothing to check");
+            }
+            return false;
+        }
+        true
+    }
+
     /// Prints information about a file in process
     pub fn print_processing_info(&self, file: &FileEntry) {
         if !self.is_quiet_mode {
--- a/src/common/output/check.rs
+++ b/src/common/output/check.rs
@@ -16,15 +16,11 @@ impl CheckOutput {
         }
     }
 
-    /// Prints information when there is nothing to check and returns a boolean
-    pub fn is_something_to_check(&self) -> bool {
-        if self.files_count == 0 {
-            if !self.is_quiet_mode {
-                println!("Nothing to check");
-            }
-            return false;
+    /// Prints a message that there is nothing to check
+    pub fn print_nothing_to_check(&self) {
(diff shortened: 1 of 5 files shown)


mstruebing (dismissed):
Great work I like it a lot :)

src/common/output/check.rs
```diff
@@ -16,6 +16,17 @@ impl CheckOutput {
...
+    /// Prints information when there is nothing to check and returns a boolean
+    pub fn is_something_to_check(&self) -> bool {
+        if self.files_count == 0 {
+            if !self.is_quiet_mode {
+                println!("Nothing to check");
+            }
+            return false;
+        }
+        true
+    }
```
mgrachev, line 28:
```suggestion
    pub fn print_nothing_to_check(&self) {
        if !self.is_quiet_mode && self.files_count == 0 {
            println!("Nothing to check");
        }
    }
```
This method should print a message and not return anything.

mgrachev (commented):
@jodli 👋 Thank you for your contribution ✌️

src/common/output/check.rs
jodli: I used the return value of this function to return early in the check function [here](https://github.com/dotenv-linter/dotenv-linter/pull/392/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R20).

This way I could use the function to notice me about an empty input list without calling the `is_empty()` method again.

DDtKey: > I used the return value of this function to return early in the check function [here](https://github.com/dotenv-linter/dotenv-linter/pull/392/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R20).
> 
> 
> 
> This way I could use the function to notice me about an empty input list without calling the `lines_map.is_empty()` method again.

Personally, I think this is not very clear logic. Because there is an implicit logic inside this method - output to the console, which is not clear from the name (`is_something_to_check`).  
In addition, I believe that this is not `Output`'s area of ​​responsibility.

jodli: You are right it's not very clear.

I will change it :)

jodli: I moved the `lines_map.is_empty()` check outside of the `Output` struct.
Now the `print_nothing_to_check()` function just prints the output.

DDtKey (approved):

mgrachev (approved):

mgrachev:
@jodli Thank you for your help ❤️

PR #393, opened 2021-03-19
Use Rc for LineEntry to reduce memory consumption

### LINKED ISSUE #388: Use Rc for LineEntry to reduce memory consumption
### Description

The `LineEntry` struct has the [FileEntry](https://github.com/dotenv-linter/dotenv-linter/blob/master/src/common/line_entry.rs#L7) field which clones for every `LineEntry` in the [get_line_entries](https://github.com/dotenv-linter/dotenv-linter/blob/master/src/lib.rs#L229) method.

Use the [Rc](https://doc.rust-lang.org/std/rc/index.html) type for `LineEntry` to reduce memory consumption.

### Implementation

**Step 1:**

Add a new method `new` for `LineEntry` which takes `file` as `Rc<FileEntry>`.

**Step 2:**

Use the method from step 1 to create the `LineEntry` struct in all places.

Tom01098:
Closes #388.

#### ✔ Checklist:

### DIFF
--- a/src/common/line_entry.rs
+++ b/src/common/line_entry.rs
@@ -1,10 +1,13 @@
-use crate::common::*;
+use std::rc::Rc;
+
 use comment::Comment;
+
+use crate::common::*;
 
 #[derive(Clone, Debug, PartialEq)]
 pub struct LineEntry {
     pub number: usize,
-    pub file: FileEntry,
+    pub file: Rc<FileEntry>,
     pub raw_string: String,
 
     /// Used in ExtraBlankLineFixer
@@ -12,6 +15,10 @@ pub struct LineEntry {
 }
 
 impl LineEntry {
+    pub fn new(number: usize, file: Rc<FileEntry>, raw_string: String, is_deleted: bool) -> Self {
+        LineEntry { number, file, raw_string, is_deleted }
+    }
+
     pub fn is_empty_or_comment(&self) -> bool {
         self.is_empty() || self.is_comment()
     }
@@ -111,6 +118,7 @@ mod tests {
 
     mod get_key {
         use super::*;
+
         #[test]
         fn empty_line_test() {
             let input = line_entry(1, 1, "");
--- a/src/common/line_entry.rs
+++ b/src/common/line_entry.rs
@@ -15,12 +15,12 @@ pub struct LineEntry {
 }
 
 impl LineEntry {
-    pub fn new(number: usize, file: Rc<FileEntry>, raw_string: String, is_deleted: bool) -> Self {
+    pub fn new(number: usize, file: Rc<FileEntry>, raw_string: String) -> Self {
         LineEntry {
             number,
             file,
             raw_string,
-            is_deleted,
+            is_deleted: false,
         }
     }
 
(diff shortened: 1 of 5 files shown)


Tom01098:
I was wondering how to enforce that `new` was called, currently anyone could just use the struct creation syntax.

I found [this](https://steveklabnik.com/writing/structure-literals-vs-constructors-in-rust) which recommended adding a private field to the struct. clippy warned against this, using [non_exhaustive](https://doc.rust-lang.org/reference/attributes/type_system.html) instead, but I found this only works at the crate level.

So what's most 'idiomatic'? I'm not sure where to go from here (but that's really irrelevant to this PR!).

src/common/line_entry.rs
```diff
@@ -1,17 +1,29 @@
...
     pub is_deleted: bool,
 }
 
 impl LineEntry {
+    pub fn new(number: usize, file: Rc<FileEntry>, raw_string: String, is_deleted: bool) -> Self {
+        LineEntry {
+            number,
+            file,
+            raw_string,
+            is_deleted,
```
mgrachev, line 23: The `is_deleted` field should be `false` by default.

src/common/line_entry.rs
```diff
@@ -1,17 +1,29 @@
...
-    pub file: FileEntry,
+    pub file: Rc<FileEntry>,
     pub raw_string: String,
 
     /// Used in ExtraBlankLineFixer
     pub is_deleted: bool,
 }
 
 impl LineEntry {
+    pub fn new(number: usize, file: Rc<FileEntry>, raw_string: String, is_deleted: bool) -> Self {
```
mgrachev, line 18:
```suggestion
    pub fn new(number: usize, file: Rc<FileEntry>, raw_string: String) -> Self {
```
The `is_deleted` field should be `false` by default.

mgrachev (commented):
@Tom01098 👋 Thank you for your contribution 🚀

I have left some comments. Please take a look at them 👀

mgrachev:
> I was wondering how to enforce that `new` was called, currently anyone could just use the struct creation syntax.
> 
> I found [this](https://steveklabnik.com/writing/structure-literals-vs-constructors-in-rust) which recommended adding a private field to the struct. clippy warned against this, using [non_exhaustive](https://doc.rust-lang.org/reference/attributes/type_system.html) instead, but I found this only works at the crate level.
> 
> So what's most 'idiomatic'? I'm not sure where to go from here (but that's really irrelevant to this PR!).

I like more the way where you can create a public `Struct` with a private field:
```rust
pub struct LineEntry {
    pub number: usize,
    pub file: Rc<FileEntry>,
    pub raw_string: String,

    /// Used in ExtraBlankLineFixer
    is_deleted: bool,
}
```

This struct you can't create like this `LineEntry { ... }` - only using the `new` method.

If you want to deal with the private field you can create a method like this:
```rust
pub fn is_deleted(&self) -> bool {
    self.is_deleted
}
```

Tom01098:
@mgrachev thanks for the review, I should have corrected everything as requested. (If only I remembered conventional commits didn't end with a full stop I wouldn't have to keep force pushing!!)

mgrachev (approved):

mgrachev:
@Tom01098 Thank you for your help ❤️

Issue #382, opened 2021-03-15
Allow unordered keys for substituion variables


DDtKey:
In variable substitution cases, `dotenv-linter` shouldn't complain on order of the keys.

Such files to be considered valid:

```env
# .env
FOO=something
BAR=$FOO # this is correct, because BAR uses FOO variable
```
For this, changes must be made to the` UnorderedKeyChecker`.

Since we already support sort groups (see the [Unordered Key documentation](https://dotenv-linter.github.io/#/checks/unordered_key)), a good way would be just consider the substitution key (which refers to another key in [current sort group](https://github.com/dotenv-linter/dotenv-linter/blob/492d3a4046fd97cdfbeb7484e430375683a06a2f/src/checks/unordered_key.rs#L6)) - as another group separator.

### TODO
Using the substitution key getter (see #379), check if this key refers to another key in the [current sort group](https://github.com/dotenv-linter/dotenv-linter/blob/492d3a4046fd97cdfbeb7484e430375683a06a2f/src/checks/unordered_key.rs#L6), and then start a new sort group [here](https://github.com/dotenv-linter/dotenv-linter/blob/492d3a4046fd97cdfbeb7484e430375683a06a2f/src/checks/unordered_key.rs#L31) if so. 

### Example

This file is incorrect because the keys are not sorted alphabetically before and after the `FOO` key (which uses `KEY`):

```env
# .env
KEY=VALUE
ABC=XYZ
FOO=$KEY
BOO=$FOO
XYZ=ABC
BAR=FOO
```

The correct version of this file should look like this:

```env
# .env
ABC=XYZ
KEY=VALUE
FOO=$KEY # Unordered, because FOO uses KEY
BAR=FOO # here starts new sort group
BOO=$FOO
XYZ=ABC
```

PR #401, opened 2021-03-24
Allow unordered keys for substitution variables (#382)

Ru5ty0ne:
Solves [#382](https://github.com/dotenv-linter/dotenv-linter/issues/382)

#### ✔ Checklist:

### DIFF
--- a/src/checks/unordered_key.rs
+++ b/src/checks/unordered_key.rs
@@ -27,14 +27,13 @@ impl Default for UnorderedKeyChecker<'_> {
 
 impl Check for UnorderedKeyChecker<'_> {
     fn run(&mut self, line: &LineEntry) -> Option<Warning> {
-        let sub_keys = line.get_substitution_keys();
-        let substitution_check = !sub_keys.is_empty()
-            && sub_keys
-                .iter()
-                .all(|&sub_key| self.keys.contains(&sub_key.to_string()));
+        let has_substitution_in_group = line
+            .get_substitution_keys()
+            .iter()
+            .any(|k| self.keys.iter().any(|key| key == k));
 
         // Support of grouping variables through blank lines and control comments
-        if line.is_empty() || line.get_control_comment().is_some() || substitution_check {
+        if line.is_empty() || line.get_control_comment().is_some() || has_substitution_in_group {
             self.keys.clear();
             return None;
         }
--- a/src/checks/unordered_key.rs
+++ b/src/checks/unordered_key.rs
@@ -222,4 +222,70 @@ mod tests {
 
         run_unordered_tests(asserts);
     }
+
+    #[test]
+    fn three_ordered_groups_with_two_unordered_substitution_keys_that_have_multiple_values_test() {
+        let asserts = vec![
+            (line_entry(1, 3, "BBB=XYZ"), None),
+            (line_entry(2, 3, "HHH=VAL"), None),
+            (line_entry(3, 3, "ZZZ=VALUE"), None),
+            (
+                line_entry(4, 3, "CCC=$NNN$HHH # Unordered, CCC uses HHH"),
+                None,
+            ),
+            (line_entry(5, 3, "BAR=FOOD"), None),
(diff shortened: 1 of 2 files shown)


src/checks/unordered_key.rs
```diff
@@ -203,4 +209,18 @@ mod tests {
 
         run_unordered_tests(asserts);
     }
+
+    #[test]
+    fn two_ordered_groups_with_two_substitution_keys_test() {
```
DDtKey, line 214: Could you add a few tests with multiple substitution keys in one variable and also with the wrong order of the keys (for example, between or before substitution keys) ?

DDtKey (commented):
Thank you for your contribution! 👍 
I left some comments, take a look at them pls 🙂

src/checks/unordered_key.rs
Ru5ty0ne: Sure, I'll do it. Thanks for feedback.

Ru5ty0ne: Done.

Ru5ty0ne: What about this case:
> CCC=$HHH
> BBB=XYZ
> HHH=VAL
> ZZZ=VALUE

Should warning tell us that CCC should be placed below HHH or leave it as is with "The BBB key should go before the CCC key"?

Ru5ty0ne: And what right order should be for this case:

> ZZZ=$AAA
> BBB=III
> AAA=VVV

1)
> AAA=VVV
> BBB=III
> ZZZ=$AAA

or

2)
> AAA=VVV
> ZZZ=$AAA
> BBB=III

DDtKey: Good questions!
I think for such cases it's not worth changing the order, because it may depend on an external variable.

I.e for this one:
```sh
CCC=$HHH
BBB=XYZ
HHH=VAL
ZZZ=VALUE
```
the correct version should look like this:
```sh
BBB=XYZ
CCC=$HHH # Uses external HHH key
HHH=VAL
ZZZ=VALUE
```

But it might be better to consider this in a separate issue 🤔

mgrachev: > the correct version should look like this:
> 
> ```shell
> BBB=XYZ
> CCC=$HHH # Uses external HHH key
> HHH=VAL
> ZZZ=VALUE
> ```

I don't fully understand why you think `$HHH` external? The next line is `HHH=VAL`.
Or I missed something?

DDtKey: > I don't fully understand why you think `$HHH` external? The next line is `HHH=VAL`.
> Or I missed something?

Because original `.env` has next order (in example above):
```shell
CCC=$HHH  # HHH can be external or not (we don't know for sure)
BBB=XYZ
HHH=VAL # this can override HHH key 
ZZZ=VALUE
```

And it is likely that the key depends on an external variable (or from another sort group).
But if we create `Warning` - this can cause confusion and for example  and a fixer will fix it.
Thus breaking the relationship between `CCC` and `HHH`

At the same time, these are complex validation rules, and in such cases, I think it would be more correct for the user to divide into groups (if necessary):

```shell
CCC=$HHH

BBB=XYZ
HHH=VAL
ZZZ=VALUE
```

And in this case, the current implementation already suits us completely.
This is controversial, so I suggested discussing it in another issue if needed.

Ru5ty0ne:
Should I do something about linter's warning about commits not following Conventional Commits format?

mgrachev:
> Should I do something about linter's warning about commits not following Conventional Commits format?

Please, rebase your earlier commits.
https://github.com/dotenv-linter/dotenv-linter/pull/401/checks?check_run_id=2211833110

Ru5ty0ne:
> Please, rebase your earlier commits.

Done.

mgrachev (approved):

mgrachev:
@Ru5ty0ne Thank you for your help 🔥

Issue #383, opened 2021-03-15
Don't fix key order for substitution variables


DDtKey:
According to #382, the `UnorderedKeyFixer` mustn't fix the order of variables if they reference to key in the same sorting group.

Such files to be considered valid and can't be fixed:
```env
# .env
FOO=something
BAR=$FOO # this is correct, because BAR uses FOO variable
```

### TODO
Using the substitution key getter (see #379), check if this key refers to another key in the current sort group ([lines](https://github.com/dotenv-linter/dotenv-linter/blob/492d3a4046fd97cdfbeb7484e430375683a06a2f/src/fixes/unordered_key.rs#L39) between [start_index](https://github.com/dotenv-linter/dotenv-linter/blob/492d3a4046fd97cdfbeb7484e430375683a06a2f/src/fixes/unordered_key.rs#L42) and `i`, where `i` - substitution variable index), and then sort the found [part](https://github.com/dotenv-linter/dotenv-linter/blob/492d3a4046fd97cdfbeb7484e430375683a06a2f/src/fixes/unordered_key.rs#L64). 
For such substitution keys need to add a comment like: `BAR=$FOO # Unordered, BAR uses FOO`.

### Example

For this file:
```env
# .env
KEY=VALUE
ABC=XYZ
FOO=$KEY
BOO=$FOO
XYZ=ABC
BAR=FOO
```
After fixing, we should get the following result:
```env
# .env
ABC=XYZ 
KEY=VALUE
FOO=$KEY # Unordered, FOO uses KEY
BAR=FOO 
BOO=$FOO 
XYZ=ABC
```
After the `FOO` key starting a new sort group and `BOO` key can be sorted, because `FOO` in another group.

PR #406, opened 2021-03-28
Don't fix key order for substitution variables (#383)

stygian-coffee:
This pull request addresses issue #383.

I think the `fix_warnings` function could still use a bit of cleanup after my changes, since I do think it nests a bit deep now.

#### ✔ Checklist:


src/fixes/unordered_key.rs
```diff
@@ -279,6 +307,40 @@ mod tests {
         assert_lines(&lines, vec!["A=B", "B=C", "D=E"]);
     }
 
+    #[test]
+    fn key_order_substitution_variable_test() {
```
mgrachev, line 311: Can you add more test cases?
For example: https://github.com/dotenv-linter/dotenv-linter/pull/401/files#diff-34a948562dd051763d9e109b86f2aad7d39f9e8ef8f072fcd674cebcfe001328R213

mgrachev (commented):
@stygian-coffee 👋 Thank you for your contribution 🚀

src/fixes/unordered_key.rs
stygian-coffee: Thank you for the comment! I have added many tests. One thing is unclear. On line 408 of `src/fixes/unordered_key.rs` I have a TODO comment because it's not clear to me whether a sort group that looks like:
```
A=1
B=2
C=$B$A
```
should result in
```
A=1
B=2
C=$B$A # Unordered, C uses A
```
or
```
A=1
B=2
C=$B$A # Unordered, C uses B
```

The test currently tests for the first case.

Let me know if you think I've missed anything!

src/fixes/unordered_key.rs
```diff
@@ -61,11 +62,38 @@ impl Fix for UnorderedKeyFixer<'_> {
...
+                    let key = match lines[j].get_key() {
+                        Some(k) => k,
+                        None => continue,
+                    };
+
+                    if subst_keys.iter().any(|k| k == &key) {
+                        replaced_line = Some(format!(
+                            "{} # Unordered, {} uses {}",
+                            &line.raw_string,
+                            &lines[i].get_key().unwrap(),
```
DDtKey, line 77: `unwrap` can cause the panic

DDtKey (commented):
Thank you for your contribution! 🚀 
I left some comments, please take a look.

src/fixes/unordered_key.rs
stygian-coffee: Wow, I didn't know that the `?` operator returns `None` in functions that return an `Option`. That's cool. I worry though that returning `None` here is a form of silently failing, and we would rather inform the user that we've reached a program state that we expected to never reach.

DDtKey: That is good point 👍 
I think it can be done at upper level by comparing number of warnings and expected number of fixes and wrap that to `Result` with custom error.

However I suppose it is worth considering in a separate issue

src/fixes/unordered_key.rs
```diff
@@ -61,11 +62,38 @@ impl Fix for UnorderedKeyFixer<'_> {
...
+                        has_substitution_variable = true;
+                        end.replace(i);
+                    }
+                }
+
+                if line.is_empty()
+                    || line.is_last_line()
+                    || is_control_comment
+                    || has_substitution_variable
+                {
```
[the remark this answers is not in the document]
stygian-coffee: I agree that this is indeed much clearer. I didn't know that `Option` implemented `IntoIterator`, so thanks for showing me the `flat_map` trick :slightly_smiling_face:.

src/fixes/unordered_key.rs
```diff
@@ -279,6 +327,236 @@ mod tests {
...
+            "A=$M",
+            "B=3",
+        ]);
+
+        let mut warnings = get_warnings(
+            &lines,
+            vec![
+                (1, "The ABC key should go before KEY key"),
+                (5, "The BAR key should go before BOO key"),
+                (9, "The A key should go before N key"),
```
mgrachev, line 385: If I run `dotenv-linter` on this file, the output is as follows:
```
UnorderedKey: The ABC key should go before the KEY key
UnorderedKey: The BAR key should go before the BOO key

Found 2 problems
```

DDtKey: Yeah, really good point 👍 
I'll put my two cents if you don't mind 🙂 

`UnorderedKeyChecker` already knows what these orders are correct (and the `Fixer` doesn't fix this in the current implementation, according to below asserts, i.e it works as expected).

I think this test should be left as is, except for this warning: 

```suggestion
```

stygian-coffee: This is indeed a mistake on my end.

src/fixes/unordered_key.rs
```diff
@@ -57,11 +57,33 @@ impl Fix for UnorderedKeyFixer<'_> {
...
 
-                if line.is_empty() || line.is_last_line() || is_control_comment {
+                if line.is_empty()
+                    || line.is_last_line()
+                    || is_control_comment
+                    || has_substitution_variables
+                {
+                    if has_substitution_variables {
+                        lines[i].raw_string =
+                            Self::with_unordered_comment(&line, substitutions_in_group)?;
```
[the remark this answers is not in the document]
stygian-coffee: I kind of like the natural sounding list of variables with the "and" in it, but okay, I have simplified it.

mgrachev (approved):

mgrachev:
@stygian-coffee Thank you for your help 🔥

PR #414, opened 2021-04-09
Add substitution key checker

### LINKED ISSUE #407: Add check: Substitution keys
Example: 

```shell
# .env

# Wrong
ABC=${BAR
FOO=$BAR}

# Correct
ABC=$BAR
FOO=${BAR}
FOO="$BAR"
```

You can find more details in #379 issue and [its implementation](https://github.com/dotenv-linter/dotenv-linter/blob/a3878621ef862fead0bf03f9344bfd45ffdad935/src/common/line_entry.rs#L92)  

And for other examples substitution keys, you can look at the [dotenv libraries that support variable substitution](https://github.com/dotenv-linter/dotenv-linter/issues/360#issuecomment-783316044).

> **The note:** 
> Read how to write a new checker on  [CONTRIBUTING.md](https://github.com/dotenv-linter/dotenv-linter/blob/master/CONTRIBUTING.md)

de-sh:
Closes #407 

#### ✔ Checklist:

- [x] Docs have been added / updated on the [dotenv-linter.github.io](https://github.com/dotenv-linter/dotenv-linter.github.io) (for bug fixes / features).

### DIFF
--- a/src/checks/substitution_key.rs
+++ b/src/checks/substitution_key.rs
@@ -1,0 +1,87 @@
+use crate::checks::Check;
+use crate::common::*;
+
+pub(crate) struct SubstitutionKeyChecker<'a> {
+    name: &'a str,
+    template: &'a str,
+}
+
+impl Default for SubstitutionKeyChecker<'_> {
+    fn default() -> Self {
+        Self {
+            name: "SubstitutionKey",
+            template: "The {} key isn't properly assigned",
+        }
+    }
+}
+
+impl Check for SubstitutionKeyChecker<'_> {
+    fn run(&mut self, line: &LineEntry) -> Option<Warning> {
+        let mut value = match line.get_value().map(str::trim) {
+            Some(value) if !value.starts_with('\'') => value,
+            _ => return None,
+        };
+
+        // Checks if keys used in value have both '{' '}' or neither
+        while let Some(index) = value.find('$') {
+            let raw_key = &value[index + 1..];
+
+            if raw_key.contains('{') ^ raw_key.contains('}') {
+                return Some(Warning::new(
+                    line.clone(),
+                    self.name,
+                    self.message(line.get_key()?),
+                ));
+            }
+
+            value = raw_key;
+        }
+        None
+    }
+
+    fn name(&self) -> &str {
+        self.name
+    }
+}
+
+impl SubstitutionKeyChecker<'_> {
+    fn message(&self, key: &str) -> String {
+        self.template.replace("{}", &key)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::common::tests::*;
+
+    fn run_substitution_tests(asserts: Vec<(LineEntry, Option<Warning>)>) {
+        let mut checker = SubstitutionKeyChecker::default();
+
+        for assert in asserts {
+            let (input, output) = assert;
+            assert_eq!(checker.run(&input), output);
+        }
+    }
+
+    #[test]
+    fn correct_substitution_key_test() {
+        let asserts = vec![
+            (line_entry(1, 3, "ABC=$BAR"), None),
+            (line_entry(2, 3, "FOO=${BAR}"), None),
+            (line_entry(3, 3, "FOO=$BAR"), None),
+        ];
+
+        run_substitution_tests(asserts);
+    }
+
+    #[test]
+    fn incorrect_substitution_key_test() {
+        let asserts = vec![
+            (line_entry(1, 2, "ABC=${BAR"), Some(Warning::new(line_entry(1, 2, "ABC=${BAR"), "SubstitutionKey", "The ABC key isn't properly assigned"))),
+            (line_entry(2, 2, "FOO=$BAR}"), Some(Warning::new(line_entry(2, 2, "FOO=$BAR}"), "SubstitutionKey", "The FOO key isn't properly assigned"))),
+        ];
+
+        run_substitution_tests(asserts);
+    }
+}
--- a/src/checks/substitution_key.rs
+++ b/src/checks/substitution_key.rs
@@ -22,11 +22,15 @@ impl Check for SubstitutionKeyChecker<'_> {
             _ => return None,
         };
 
+        let is_escaped =
+            |prefix: &str| prefix.chars().rev().take_while(|ch| *ch == '\\').count() % 2 == 1;
+
         // Checks if keys used in value have both '{' '}' or neither
         while let Some(index) = value.find('$') {
+            let prefix = &value[..index];
             let raw_key = &value[index + 1..];
 
-            if raw_key.contains('{') ^ raw_key.contains('}') {
+            if !is_escaped(prefix) && raw_key.contains('{') ^ raw_key.contains('}') {
                 return Some(Warning::new(
                     line.clone(),
                     self.name,
@@ -34,7 +38,7 @@ impl Check for SubstitutionKeyChecker<'_> {
                 ));
             }
 
-            value = raw_key;
+            value = &raw_key;
         }
         None
     }
@@ -69,7 +73,7 @@ mod tests {
         let asserts = vec![
             (line_entry(1, 3, "ABC=$BAR"), None),
             (line_entry(2, 3, "FOO=${BAR}"), None),
-            (line_entry(3, 3, "FOO=$BAR"), None),
+            (line_entry(3, 3, "FOO=\"$BAR\""), None),
         ];
 
         run_substitution_tests(asserts);
@@ -94,6 +98,18 @@ mod tests {
                     "The FOO key isn't properly assigned",
                 )),
             ),
+        ];
+
+        run_substitution_tests(asserts);
+    }
+
+    #[test]
+    fn escaped_incorrect_substitution_key_test() {
+        let asserts = vec![
+            (line_entry(1, 4, "ABC=\\${BAR"), None),
+            (line_entry(2, 4, "FOO=\\$BAR}"), None),
+            (line_entry(3, 4, "FOO=\"\\${BAR\""), None),
+            (line_entry(4, 4, "FOO=\"\\$BAR}"), None),
         ];
 
         run_substitution_tests(asserts);
--- a/src/checks/substitution_key.rs
+++ b/src/checks/substitution_key.rs
@@ -104,6 +104,41 @@ mod tests {
     }
 
     #[test]
+    fn multiple_substitution_key_test() {
+        let asserts = vec![
+            (line_entry(1, 3, "ABC=${BAR}$XYZ"), None),
+            (line_entry(2, 3, "FOO=$ABC${BAR}"), None),
+            (line_entry(3, 3, "BIZ=$FOO-$ABC"), None),
+        ];
+
+        run_substitution_tests(asserts);
+    }
+
+    #[test]
+    fn incorrect_multiple_substitution_key_test() {
+        let asserts = vec![
+            (
+                line_entry(1, 2, "ABC=${BAR$XYZ}"),
+                Some(Warning::new(
+                    line_entry(1, 2, "ABC=${BAR$XYZ}"),
+                    "SubstitutionKey",
+                    "The ABC key is not assigned properly",
+                )),
+            ),
+            (
+                line_entry(2, 2, "FOO=${ABC-$BAR}"),
+                Some(Warning::new(
+                    line_entry(2, 2, "FOO=${ABC-$BAR}"),
+                    "SubstitutionKey",
+                    "The FOO key is not assigned properly",
+                )),
+            ),
+        ];
+
+        run_substitution_tests(asserts);
+    }
+
+    #[test]
     fn escaped_incorrect_substitution_key_test() {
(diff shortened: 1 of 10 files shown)


src/checks/substitution_key.rs
```diff
@@ -0,0 +1,87 @@
...
+            assert_eq!(checker.run(&input), output);
+        }
+    }
+
+    #[test]
+    fn correct_substitution_key_test() {
+        let asserts = vec![
+            (line_entry(1, 3, "ABC=$BAR"), None),
+            (line_entry(2, 3, "FOO=${BAR}"), None),
+            (line_entry(3, 3, "FOO=$BAR"), None),
```
DDtKey, line 72:
```suggestion
            (line_entry(3, 3, "FOO=\"$BAR\""), None),
```

src/checks/substitution_key.rs
```diff
@@ -0,0 +1,101 @@
...
+
+impl Check for SubstitutionKeyChecker<'_> {
+    fn run(&mut self, line: &LineEntry) -> Option<Warning> {
+        let mut value = match line.get_value().map(str::trim) {
+            Some(value) if !value.starts_with('\'') => value,
+            _ => return None,
+        };
+
+        // Checks if keys used in value have both '{' '}' or neither
+        while let Some(index) = value.find('$') {
```
DDtKey, line 26: We need to make sure it's not an escaped character ( `$`)
Take a look at the [logic for getting substitution keys.](https://github.com/dotenv-linter/dotenv-linter/blob/a603ddab6d4d04a488fc2387478af35a795be1f4/src/common/line_entry.rs#L100)

If a character is escaped we shouldn't consider it  🙂 
Also add tests for these cases please

DDtKey (commented):
Thank you for PR 🚀 
I left a couple of comments, please take a look 🙂

src/checks/substitution_key.rs
de-sh: Do you think the latest commit solves this? 😄

de-sh:
Please forgive me for the force pushes, I am not familiar with the git commit messaging convention 😅

mgrachev (commented):
@de-sh 👋 Thank you for your contribution 🚀

I have left some comments. Please take a look at them 👀

mgrachev:
Also, please do this one:
> - [ ] Docs have been added / updated on the [dotenv-linter.github.io](https://github.com/dotenv-linter/dotenv-linter.github.io) (for bug fixes / features).

de-sh:
I did some thinking and was slightly intrigued by the following:
```bash
ABC=$B{AR
```
Should we be concerned with the `{}`s matching in such strings?

DDtKey:
I apologize for the long answer.
I just remembered and wanted to write about the fact that values can contain several substitution keys and it needs to be covered with tests. 
So good point, thanks for your vigilance @de-sh 🙂 

In your example:
```shell
ABC=$B{AR # Could be a potential Warning. But I think that `$B` should be considered the correct option. I only think of `{AR`
FOO=$BAR${ABC} # this is definitely the correct version of the substitution keys (`$BAR` and `${ABC}`)
XYZ=${FOO${BAR} # it is Warning. ${FOO does not end with`}` until the next `$` and `{`, and we should find it out
``` 

I think the algorithm should be similar to the [get_substitution_keys](https://github.com/dotenv-linter/dotenv-linter/blob/a603ddab6d4d04a488fc2387478af35a795be1f4/src/common/line_entry.rs#L92) method. Only in case of any invalid substitution key inside the value - we need to return `Warning` 🤔

DDtKey:
@ de-sh in accordance with [my comment above](https://github.com/dotenv-linter/dotenv-linter/pull/414#issuecomment-818010470), can you add tests with multiple substitution keys inside the value? 
Also as I wrote `XYZ=${FOO${BAR}` should be considered a `Warning`, please make this change  🙂 

If you have any questions, I will be glad to answer!

de-sh:
I tried some stuff with the following, but still unable to get the proper output for `${FOO${BAR}` still. Any ideas?

```rust
let is_improper_key = |key: &str| key.chars().any(|c: char| !c.is_ascii_alphanumeric() && c != '_');
```

DDtKey:
> I tried some stuff with the following, but still unable to get the proper output for `${FOO${BAR}` still. Any ideas?
> 
> ```rust
> let is_improper_key = |key: &str| key.chars().any(|c: char| !c.is_ascii_alphanumeric() && c != '_');
> ```

Alternatively, you can pre-clear the `raw_key` from excess (if exists), for example something like this:
```rust
let raw_key = &value[index + 1..];
let raw_key = raw_key.find(|c: char| c == '$' || c == '}')
        .map(|i| &raw_key[..i + 1])
        .unwrap_or_else(|| raw_key);
```

This solution may have some edge cases and may not be ideal, but at first glance it solves the problem  🤔

src/checks/substitution_key.rs
```diff
@@ -0,0 +1,152 @@
...
+            _ => return None,
+        };
+
+        let is_escaped =
+            |prefix: &str| prefix.chars().rev().take_while(|ch| *ch == '\\').count() % 2 == 1;
+
+        // Checks if keys used in value have both '{' '}' or neither
+        while let Some(index) = value.find('$') {
+            let prefix = &value[..index];
+            let raw_key = &value[index + 1..];
```
DDtKey, line 31: For clarity, what I meant in [the comment ](https://github.com/dotenv-linter/dotenv-linter/pull/414#issuecomment-821870888)
```suggestion
            let raw_key = &value[index + 1..];
            let raw_key = raw_key
                .find(|c: char| c == '$' || c == '}')
                .map(|i| &raw_key[..i + 1])
                .unwrap_or_else(|| raw_key);
```

de-sh:
Would this be enough? Would need help reducing the code: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4777e0382e67c32ce9cf9c2ab4eb8dbd

src/checks/substitution_key.rs
```diff
@@ -0,0 +1,184 @@
...
+            (
+                line_entry(1, 2, "ABC=${BAR"),
+                Some(Warning::new(
+                    line_entry(1, 2, "ABC=${BAR"),
+                    "SubstitutionKey",
+                    "The ABC key is not assigned properly",
+                )),
+            ),
+            (
+                line_entry(2, 2, "FOO=${BAR!}"),
```
DDtKey, line 118: It's correct substitution key 🙂 
There can be can be non-ascii-alphanumeric characters between `{` and `}` (exclude only `$`)

src/checks/substitution_key.rs
```diff
@@ -0,0 +1,184 @@
...
+
+        // Checks if keys used in value have both '{' '}' or neither
+        while let Some(index) = value.find('$') {
+            let prefix = &value[..index];
+            let raw_key = &value[index + 1..];
+
+            // Separate initial key from the rest
+            let (initial_key, rest) = raw_key
+                .find(|c: char| c == '$')
+                .map(|i| (&raw_key[..i], &raw_key[i..]))
```
DDtKey, line 36:
```suggestion
                .find('$')
                .map(|i| raw_key.split_at(i))
```

src/checks/substitution_key.rs
```diff
@@ -0,0 +1,184 @@
...
+        let asserts = vec![
+            (
+                line_entry(1, 2, "ABC=${BAR"),
+                Some(Warning::new(
+                    line_entry(1, 2, "ABC=${BAR"),
+                    "SubstitutionKey",
+                    "The ABC key is not assigned properly",
+                )),
+            ),
+            (
```
DDtKey, line 117: Also, we should cosider `ABC=$BAR}` as invalid substitution key (because they potentially forgot to specify `{`).
Please, add this case 🙂 

```suggestion
            ),
            (
                line_entry(2, 2, "FOO=$BAR}"),
                Some(Warning::new(
                    line_entry(2, 2, "FOO=$BAR}"),
                    "SubstitutionKey",
                    "The FOO key is not assigned properly",
                )),
            ),
            (
```

src/checks/substitution_key.rs
```diff
@@ -0,0 +1,184 @@
...
+                    .find(|c: char| !c.is_ascii_alphanumeric() && c != '_')
+                    .map(|i| (&initial_key[..i], false))
+                    .unwrap_or_else(|| (initial_key, false)),
+            };
+
+            if (bad_substitution
+                || key
+                    .chars()
+                    .any(|c: char| !c.is_ascii_alphanumeric() && c != '_'))
+                && !is_escaped(prefix)
```
DDtKey, line 56: I think this can be minimized since the value has already been pre-truncated 🤔 

```suggestion
            let has_start_brace = initial_key.starts_with('{');
            let has_end_brace = initial_key.contains('}');
            let is_incorrect_substitution = has_start_brace ^ has_end_brace;

            if is_incorrect_substitution && !is_escaped(prefix) {
```

DDtKey (commented):
I left a couple of comments, please take a look

src/checks/substitution_key.rs
de-sh: `!@$^&*(),;<>` and some more also fail, right?

src/checks/substitution_key.rs
de-sh: That is supposed to work I believe

src/checks/substitution_key.rs
de-sh: Ok, seems like that would be nearly the same as before last commit though, but will fail in some of the cases.

src/checks/substitution_key.rs
DDtKey: My mistake, you are right 🙂 
In this case, in the option I proposed, something needs to be changed (but not a lot)

src/checks/substitution_key.rs
DDtKey: > That is supposed to work I believe

With the approach I proposed below, it definitely works. But not in the current one, if I'm not mistaken

src/checks/substitution_key.rs
DDtKey: > Ok, seems like that would be nearly the same as before last commit though, but will fail in some of the cases.

In general, yes, but in previous versions we did not truncate the value to the next `$`.

DDtKey, line 56: Given [this moment](https://github.com/dotenv-linter/dotenv-linter/pull/414/files/9f18381ca71f8296fd49c6a89adc07930c3bda07#r615408818), you can reduce the code somehow like this:
```suggestion
            let end_brace_index = initial_key.find('}');

            let has_start_brace = initial_key.starts_with('{');
            let has_end_brace = end_brace_index.is_some();

            let is_incorrect_substitution = has_start_brace ^ has_end_brace
                || end_brace_index 
                    .map(|i| &initial_key[1..i])
                    .filter(|key| key.contains(|c: char| !c.is_ascii_alphanumeric() && c != '_'))
                    .is_some();

            if is_incorrect_substitution && !is_escaped(prefix) {
```

DDtKey: Take a look at the [slightly updated version](https://github.com/dotenv-linter/dotenv-linter/pull/414/files#r615419849)

de-sh: Done :+1:

src/checks/substitution_key.rs
de-sh: I meant to say that in bash, it is such:
```bash
$ BAR="hello"
$ FOO=$BAR}
$ echo $FOO
hello}
```

DDtKey: In this case, the linter's task is to identify even potential warnings.  So as described in the ischew, I would rather consider this as `Warning` (as described in issue) 🤔

src/checks/substitution_key.rs
```diff
@@ -0,0 +1,177 @@
...
+            |prefix: &str| prefix.chars().rev().take_while(|ch| *ch == '\\').count() % 2 == 1;
+
+        // Checks if keys used in value have both '{' '}' or neither
+        while let Some(index) = value.find('$') {
+            let prefix = &value[..index];
+            let raw_key = &value[index + 1..];
+
+            // Separate initial key from the rest
+            let (initial_key, rest) = raw_key
+                .find(|c: char| c == '$')
```
DDtKey, line 35: Here you can simplify the predicate
```suggestion
                .find('$')
```

src/checks/substitution_key.rs
```diff
@@ -0,0 +1,177 @@
...
+                .map(|i| raw_key.split_at(i))
+                .unwrap_or_else(|| (raw_key, ""));
+
+            // Ensure if key starts with a '{' that it ends with a '}', else set
+            // bad_substitution to true. Also set key to the proper substitution key.
```
DDtKey, line 40: Actualize this comments(or remove them), please

DDtKey (approved):
You did a great job, thank you very much! 🚀

de-sh:
Thank you for the guidance and support! ❤️

mgrachev (approved):

mgrachev:
@de-sh Awesome 🔥 Thank you for your help ❤️

Issue #410, opened 2021-04-02
Use Rc for Warning to reduce memory consumption


mgrachev:
### Description

The `Warning` struct has the [LineEntry](https://github.com/dotenv-linter/dotenv-linter/blob/master/src/common/warning.rs#L8) field which clones in every checks, e.g.: [DuplicatedKeyChecker](https://github.com/dotenv-linter/dotenv-linter/blob/master/src/checks/duplicated_key.rs#L32).

Use the [Rc](https://doc.rust-lang.org/std/rc/index.html) type for `Warning` to reduce memory consumption.

Similar issue: https://github.com/dotenv-linter/dotenv-linter/issues/388

PatatasDelPapa:
I want to do this one but is my first time trying to contribute to a project and I'm a bit lost.
For now I forked the repository and run the tests as it is. 1 failed like this

```markdown
failures:

---- fixes::fixtures stdout ----
thread 'fixes::fixtures' panicked at 'assertion failed: `(left == right)`
  left: `"\n# dotenv-linter:off UnorderedKey\n_ENV=000\nA=B\nA_ENV_11=B\nA_ENV_12=C\nA_ENV_13=D\nA_ENV_14=123\nA_ENV_15=345\nAB_ENV_16=777\nAB_ENV_17=alpha\nAB_ENV_18=beta\nAB_ENV_19=alpha12\nAB_ENV_20=beta13\nB=C\nC=D\nD=123\n# dotenv-linter:off SpaceCharacter\nE=345\nENV11=C\n# A_ENV_12=C\nSOME_ENV_WITHOUT_EQ=\nANOTHER_ENV=\nENV12_=D\n# dotenv-linter:on UnorderedKey\n\n# comment\n\n# Another\n# multiline\n# comment\n# ANOTHER_ENV=\n# A_ENV_12=C\nENV10=B\n# ENV11=C\n# ENV12_=D\n# single line comment\nENV13=123\nENV14=345\n# Multiline\n# comment\nSOME_ENV=\nVAR_1 = 1234\nVAR_2=1234\nVAR_3=1234\nVAR_4=1234\n# dotenv-linter:on SpaceCharacter\n# AB_ENV_19=alpha14\n# 
[...]

PR #421, opened 2021-04-19
Refactor Check unit tests

### LINKED ISSUE #419: Refactor Check impl unit testing
Hello 👋 
I found this repo from "This Week in Rust", I'm new and looking to gain some experience with contributing to Rust projects! 

I've noticed that the unit tests for Check implementations are all very similar so contain lots of boiler plate. I feel this could be a pain point when changing the API of the Check::run method, which will probably occur in resolving #410 and possibly in the future?

https://github.com/dotenv-linter/dotenv-linter/blob/5d4e8ed547c06de6d9266557a8cfed57c03dd185/src/checks/extra_blank_line.rs#L55-L66

The above is a great example of how extra_blank_line mod refactors out most of the boiler plate for each test! 

I think extracting this common function would make unit testing easier; also changes to Check, LineEntry, Warning, won't require many small changes to many unit tests (in respect to Check impls) but just to the core/common 'run_assert' like function. 

A macro could further lessen the boilerplate and I think the following would be possible:

```rust
check_tester! {
    DuplicatedKeyChecker;
    with_one_duplicated_key_test => {
        "FOO=BAR" => None,
        "FOO=BAR" => Some("The FOO key is duplicated"),
    },
    with_two_unique_keys_test => {
        "FOO=BAR" => None,
        "BAR=FOO" => None,
    },
}
```
This would be equivalent to the two test functions found in the duplicated_key mod (below):
https://github.com/dotenv-linter/dotenv-linter/blob/5d4e8ed547c06de6d9266557a8cfed57c03dd185/src/checks/duplicated_key.rs#L49-L83

I'm not sure whether doing this would however, lessen the experience for new contributors as they don't need to roll out their own unit tests from scratch? 

Please let me know what you think about this idea! I'm happy to do the work for this but wanted to check with this issue first.

mc1098:
Refactors the unit tests for Check implementations as per #419. 

Unit tests have been replaced with a new check_tester macro which takes care of the boiler plate of writing multiple test cases. This would make the tests easier to change when an interface change is required, such as #410, as only the expanded code of the macro is changed to reflect the new API.  

This change would also effect #414 and once that PR is merged I'm happy to make the changes so that it is included in this refactoring :)

Closes #419

#### ✔ Checklist:

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -26,6 +26,34 @@ pub(crate) mod tests {
     use std::path::PathBuf;
     use std::rc::Rc;
 
+    #[macro_export]
+    macro_rules! check_tester {
+        (@token $t:tt $sub:expr) => {$sub};
+        (@count $($t:tt)*) => {<[()]>::len(&[$(check_tester!(@token $t ())),*])};
+        ($checker:ident;
+            $(
+                $test:ident => {$(
+                    $input:expr => $expected:expr,
+                )*}
+            ),* $(,)?) => {
+            $(
+                #[test]
+                fn $test() {
+                    let mut checker = $checker::default();
+                    let total = check_tester!(@count $($expected)*);
+                    let mut _line_number = 1;
+                    $(
+                        let line = line_entry(_line_number, total, $input);
+                        _line_number += 1;
+                        let result = checker.run(&line);
+                        let expected = ($expected).map(|e: &str| Warning::new(line, checker.name(), e));
+                        assert_eq!(expected, result);
+                    )*
+                }
+            )*
+        };
+    }
+
     #[allow(dead_code)]
     pub fn blank_line_entry(number: usize, total_lines: usize) -> LineEntry {
         LineEntry::new(
--- a/src/common.rs
+++ b/src/common.rs
@@ -29,10 +29,10 @@ pub(crate) mod tests {
     use std::rc::Rc;
 
     /**
-        Helper function for testing `Check` implementations. 
+        Helper function for testing `Check` implementations.
 
         A `Check` implementation can be used against a number of &str inputs
-        and optional `Warning` messages respectively. 
+        and optional `Warning` messages respectively.
 
         This function construct `LineEntry`s and optional `Warning`s,
         if required, in order to assert that the `Check` implementation is creating
@@ -55,7 +55,7 @@ pub(crate) mod tests {
         any warnings, hence the `None`.
 
         The second line however, should expect a `Warning` with a message of
-        "The FOO key is duplicated". 
+        "The FOO key is duplicated".
     */
     pub fn check_test<'test, T, U>(checker: &mut T, asserts: U)
     where
--- a/src/common.rs
+++ b/src/common.rs
@@ -26,6 +26,75 @@ pub(crate) mod tests {
     use std::path::PathBuf;
     use std::rc::Rc;
 
+    /**
+       Creates test cases for Check trait implementations.
+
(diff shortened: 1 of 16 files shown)


mc1098:
I've just updated this PR to take into account the recently merged #414 🎉

mgrachev:
@mc1098 👋 Thank you for your contribution 👍

Sorry, I'm quite busy now, so I'll try to review your PR by the end of the week 🙁

src/common.rs
```diff
@@ -26,6 +26,103 @@ pub(crate) mod tests {
...
+                        _line_number += 1;
+                        let result = checker.run(&line);
+                        let expected = ($expected).map(|e: &str| Warning::new(line, checker.name(), e));
+                        assert_eq!(expected, result);
+                    )*
+                }
+            )*
+        };
+    }
+
```
DDtKey, line 125: In my opinion, declarative macros make a big contract here and are more difficult to understand and maintain in the future 🤔 
But I like going in this direction, I would prefer to use a generalized generic function, and reuse it in all tests in the same way.
Something like that:
```rust
    pub fn test_check<T>(checker: &mut T, asserts: Vec<(&str, Option<&str>)>) where T: Check {
        let mut line_number = 1;
        let total = asserts.len();

        for (input, expected) in asserts {
            let line = line_entry(line_number, total, input);
            line_number += 1;

            let result = checker.run(&line);
            let expected = expected.map(|e: &str| Warning::new(line, checker.name(), e));

            assert_eq!(result, expected);
        }
    }

```

As a result, the tests will look something like this:
```rust
    // UnorderedKeyChecker tests
    #[test]
    fn one_unordered_key_test() {
        test_check(&mut UnorderedKeyChecker::default(), vec![
            ("FOO=BAR", None),
            ("BAR=FOO", Some("The BAR key should go before the FOO key")),
        ]);
    }
    
    // QuoteCharacterChecker tests
    const WARNING: &str = "The value has quote characters (\', \")";

    #[test]
    fn with_single_quote_test() {
        test_check(&mut QuoteCharacterChecker::default(), vec![
            ("FOO=BAR", None),
            ("FOO='BAR'", Some(WARNING)),
            ("FOO='B\"AR'", Some(WARNING)),
            ("FOO=\'BAR BAR\'", None),
        ]);
    }
```

What do you think of this @mc1098  @dotenv-linter/core ?

Perhaps in this case a little more code, but it seems to me more clear and separated (no explicit contract with custom syntax)

mc1098: @DDtKey Thank you 😃!
I'm happy with either macro or function but I agree that the macro here makes it more complicated to change in the future. A function might be bit more code per test but is also more explicit, you can infer what that function is probably doing by it's signature but the same isn't quite true with the macro as it's bit more magic. 
I shall update this PR maybe today or tomorrow to use a generic function instead 🚀

mc1098: I have pretty much copied your suggestion. I changed the Vec for AsRef slice, which I think makes it more ergonomic and would avoid requiring allocation when called with an array over a Vec.  
https://github.com/dotenv-linter/dotenv-linter/blob/5fc4b217712428b9c4df9b1fe5878ae0189664c1/src/common.rs#L60-L63

DDtKey (approved):
Thank you for the great job! 🚀

mgrachev (dismissed):

mgrachev (dismissed):

mgrachev (approved):

mgrachev:
@mc1098 Thank you for your help 🔥

PR #422, opened 2021-04-29
Fixer tests refactoring

atsmtat:
There are two major refactoring done in this PR, both for the unit tests in fixers. This PR is also a preparation for #410 
1.  Introduced a new macro `lines_and_warnings` which makes it easier to define input for fixer tests. Fixers work with lines and warnings, and so the tests require to prepare a `Vec<LineEntry>` and a `Vec<Warning>`. Using this macro, we avoid writing boilerplate code. It also makes it easier to change `LineEntry` or `Warning` interface in future (which will happen in #410).
2. Added a new generic function `run_fix_warnings`, which takes any `Fix` object, and runs its `fix_warnings` over the lines and warnings passed in. This abstracts away the details of how lines and warnings are passed to `fix_warnings`. Towards that, it takes the ownership of both lines and warnings vector, so that they can be morphed into appropriate input for `fix_warnings` call. Fix for #410 would require changing the signature of `fix_warnings`.

NOTE: I haven't committed changes converting unordered_key tests to use new `lines_and_warnings` macro, as it has total 22 tests. I'd do it once I get your feedback on this new macro.

#### ✔ Checklist:

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -110,6 +110,40 @@ pub(crate) mod tests {
         assert_eq!("FOO-BAR", remove_invalid_leading_chars(string));
     }
 
+    /**
+    The purpose of this macro is to avoid boilerplate code while defining
+    input for tests in `Fix` trait implementations.
+    Since fixers work on a sequence of input lines and a set of warnings,
+    these tests often have to prepare a `Vec<LineEntry>` and a `Vec<Warning>`
+    to pass it to `Fix` methods.
+
+    Using this macro, input lines and warnings can be encoded in a vector
+    format, where each entry is a pair of a line and an optional warning
+    content. The macro takes care of determining the line numbers, thus
+    avoiding hard-coded integers.
+
+    Macro expands to a block whose value is of type
+    `(Vec<LineEntry>, Vec<Warning>)`.
+
+    # Examples:
+    ```
+    // input with 3 lines, 1 warning
+    let (lines, warnings) = lines_and_warnings![
+        "foO=BAR" => Some(("LowercaseKey","The FOO key should be in uppercase")),
+        "Z=Y" => None,
+        "" => None,
+    ];
+    ```
+
+    ```
+    // input with 3 lines, 0 warning
+    let (lines, warnings) = lines_and_warnings![
+        "FOO=BAR" => None,
+        "Z=Y" => None,
+        "" => None,
+    ];
+    ```
+    */
     #[cfg(test)]
     #[macro_export]
     macro_rules! lines_and_warnings {
--- a/src/common.rs
+++ b/src/common.rs
@@ -109,4 +109,27 @@ pub(crate) mod tests {
         let string = "***FOO-BAR";
         assert_eq!("FOO-BAR", remove_invalid_leading_chars(string));
     }
+
+    #[cfg(test)]
+    #[macro_export]
+    macro_rules! lines_and_warnings {
+        (
+            $(
+                // each repeat must contain `expr => expr`
+                $line:expr => $opt_msg:expr
+            ),* $(,)*
+            // ...zero or more, separated by commas
+        ) => {
+            // replace with multi-line statment block
+            {
+                let lines = vec![ $( $line ),* ];
+                let warnings = vec![ $( $opt_msg ),* ];
+                let line_warnings: Vec<(usize, &str)> = warnings
+                    .iter()
+                    .enumerate()
+                    .filter_map(|(i, opt_msg)| opt_msg.and_then(|msg| Some((i, msg)) )).collect();
+                (lines, line_warnings)
+            }
+        };
+    }
 }
--- a/src/common.rs
+++ b/src/common.rs
@@ -116,19 +116,32 @@ pub(crate) mod tests {
         (
             $(
                 // each repeat must contain `expr => expr`
-                $line:expr => $opt_msg:expr
+                $line:expr => $opt_warning:expr
             ),* $(,)*
             // ...zero or more, separated by commas
         ) => {
             // replace with multi-line statment block
             {
-                let lines = vec![ $( $line ),* ];
-                let warnings = vec![ $( $opt_msg ),* ];
-                let line_warnings: Vec<(usize, &str)> = warnings
+                let lines_input = vec![ $( $line ),* ];
+                let warnings_input = vec![ $( $opt_warning ),* ];
+                let total_lines = lines_input.len();
+
+                let line_entries: Vec<LineEntry> = lines_input
                     .iter()
                     .enumerate()
-                    .filter_map(|(i, opt_msg)| opt_msg.and_then(|msg| Some((i, msg)) )).collect();
-                (lines, line_warnings)
+                    .map(|(i, content)| line_entry(i + 1, total_lines, content))
+                    .collect();
+
+                let warnings: Vec<Warning> = warnings_input
+                    .iter()
+                    .enumerate()
+                    .filter_map(|(i, opt_warn)| {
+                        opt_warn.and_then(|(kind, msg): (&str, &str)| {
+                            Some(Warning::new(line_entries[i].clone(), kind, msg))
+                        })
+                    })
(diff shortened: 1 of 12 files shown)


src/fixes.rs
```diff
@@ -94,48 +94,57 @@ pub fn run(warnings: &mut [Warning], lines: &mut Vec<LineEntry>, skip_checks: &[
...
+    mut lines: Vec<LineEntry>,
+    mut warnings: Vec<Warning>,
+) -> (Option<usize>, Vec<String>) {
+    let warnings_mut = warnings.iter_mut().collect();
+    let fix_count = fixer.fix_warnings(warnings_mut, &mut lines);
+
+    // Remove lines marked as deleted
+    lines.retain(|l| !l.is_deleted);
+
+    let fixed_lines: Vec<String> = lines.iter().map(|le| le.raw_string.clone()).collect();
```
DDtKey, line 109: As I can see, we still use cloning (but for strings), which is what we want to avoid in #410
Is it possible to avoid this directly in the linter (not in tests)? Otherwise, I would have preferred `Rc` 🤔

src/common.rs
```diff
@@ -60,4 +60,74 @@ pub(crate) mod tests {
...
+    let (lines, warnings) = lines_and_warnings![
+        "FOO=BAR" => None,
+        "Z=Y" => None,
+        "" => None,
+    ];
+    ```
+    */
+    #[cfg(test)]
+    #[macro_export]
+    macro_rules! lines_and_warnings {
```
DDtKey, line 100: Please pay attention to the [discussion in parallel PR about the macro](https://github.com/dotenv-linter/dotenv-linter/pull/421#discussion_r622391290)
In my opinion, we can go the same way here too 🙂

DDtKey (commented):
Thank you for your contribution 👍

atsmtat:
> Thank you for your contribution 👍

atsmtat:
Sorry, I closed the PR by mistake.

src/fixes.rs
atsmtat: Thanks for the review @DDtKey! 🙂 

This function is only for testing purposes, so I thought it should be OK to trade performance with the ease of use. Let me know if you think otherwise.

As for fixers in production code, here's my proposal as a fix for #410 --

I noticed that `fix_warnings` take mutable warnings -- `Vec<&mut Warning>`, but the implementations only need to know line numbers which require the fix. Current implementations are just grabbing `line_number` out of `Warning`, and in one case, warnings are simply discarded. So, I was considering changing `fix_warnings` to take `fix_lines: Vec<usize>`, i.e. line numbers instead of warnings. The caller `fixes::run` would also be changed to not take `warnings: &mut [Warning]`, instead it would take a map of warning name to line numbers (this would also fix the current comment "// We can optimize it: create check_name:warnings map in advance" in it).

Once fixers are no longer using mutable warnings (as they don't really have to), we can change `Warning` to hold a reference to `LineEntry` instead of a copy. Once the checkers are done running, we go over the warnings generated, and prepare the map I mentioned, and pass mutable lines along with this map to the `fixes::run`.

The major downside I see in using `Rc` to manage `LineEntry`s is that we'd loose the contiguous storage of `LineEnty`s in the memory, as it would be `Vec<Rc<LineEntry>>` instead of `Vec<LineEntry>`. Individual `LineEntry` objects would be scattered across the heap, which is not as good for the performance as having them in contiguous array (due to cache locality benefits).

What do you think of this @DDtKey? If you think my proposal sounds complicated, or if you have plans of adding fixers requiring more than just line numbers, let me know. I'd be happy to go with `Rc` if you still prefer it 🙂

src/common.rs
atsmtat: Ah my bad, didn't notice. I could convert this into a function like you suggested in the other PR. Will update this PR soon.

src/fixes.rs
DDtKey: In tests, this is acceptable, I agree 🙂 

On the subject of lifetimes:  it's still difficult for me to present your proposal based on the description 🤔 
Since before passing them to `fix_warnings`, they are filtered for pass to the corresponding fixer (currently by name).
Could you provide some function signatures for demo or even _proof of concept_ (e.g. later in a separate PR) ?

atsmtat: I opened #423 with _proof of concept_. Let me know what you think 🙂

src/common.rs
atsmtat: I updated the PR with new changes which no longer use macro 🙂

mgrachev:
@atsmtat Thank you for your contribution ✌️

To tell the truth, your solution looks pretty complicated. Could you please simplify it?

I think you can do something similar like in this PR https://github.com/dotenv-linter/dotenv-linter/pull/421/files#diff-e24518d69b6357926bce35868225330c1992a4029b521c0112b4ec8523ddc31cR60
Just add one generic function for use in tests.

mgrachev:
@atsmtat Hey 👋 Do you have any progress on this?

mgrachev:
@atsmtat I'm sorry, but I'll close the PR.

When you are ready to finish this PR, please reopen it.

atsmtat:
> @atsmtat Thank you for your contribution ✌️
> 
> To tell the truth, your solution looks pretty complicated. Could you please simplify it?
> I think you can do something similar like in this PR https://github.com/dotenv-linter/dotenv-linter/pull/421/files#diff-e24518d69b6357926bce35868225330c1992a4029b521c0112b4ec8523ddc31cR60
> Just add one generic function for use in tests.

@mgrachev I agree that new solution is a bit involved. Main difference between the checker test solution that you pointed out and fixer test is the test input. Checker interface is simple. Input is a set of lines, output is optional warning(s) per input line. So, test (input, output) is compact -- `[ (line, optional warning) ]`. (Although it's possible for single line to have multiple warnings, I didn't find tests practicing it. Otherwise you'd need `[ (line, [ warning ]) ]`)

Fixer interface is a bit complicated. Input is a set of `(line, zero or more warnings)` pairs, and output is `(fixed_count, fixed_lines)`. In my solution, I decided to use `TestLine` struct to represent an input line with warning(s). It uses builder pattern so that you can append multiple warnings to the same line via `warning` call. Then the test input becomes a collection of such `TestLine`s -- `TestLineEntries` (which comes with a handy method to split entries into `[lines]` and `[warnings]` to pass on to the fixer runner). If we were to avoid having these helper structs, input vector [here](https://github.com/dotenv-linter/dotenv-linter/pull/422/files#diff-d47f5ff4076781433909302c51fe0c6a1495f264ac6928e29a519cab851d7997R72) would look something like this:

```rust
            vec![
                ("FOO=BAR", []),
                ("Z=Y", []),
                ("FOO=BAZ", [("Duplicatedkey", "The Foo key is duplicated")]),
                ("Z=X", [("Duplicatedkey", "The Z key is duplicated")]),
            ]
```

If you were to test a line with two warnings, where only one of them should be fixed by a given fixer, it would look like:

```rust
("z=X", [("Duplicatedkey", "The z key is duplicated"), ("LowercaseKey","The z key should be in uppercase")])
```

I've added a new test practicing multi-warning [here](https://github.com/dotenv-linter/dotenv-linter/pull/422/files#diff-9da81e837fc18deb16c23f50802ca4733447d59623874056ffb7d84522030ed7R125), to see how it looks like in my solution.

Also, the kind of generic function you pointed out is also present in my solution, [here](https://github.com/dotenv-linter/dotenv-linter/pull/422/files#diff-e24518d69b6357926bce35868225330c1992a4029b521c0112b4ec8523ddc31cR198). But it's not single-funtion-to-do-it-all because of the nature of fixer interface.

If you prefer using a combination of tuples and slices over helper structs, I'd be happy to make the change :)

atsmtat:
> @atsmtat I'm sorry, but I'll close the PR.
> 
> When you are ready to finish this PR, please reopen it.

Sorry, I got busy with something. I've replied to your earlier concern. Once I hear from you, I'll update the PR.

PR #423, opened 2021-05-05
Avoid making a copy of LineEntrys for warnings

atsmtat:
This is a _proof of concept_ for the fix for #410 

Store a borrowed reference instead of a copy of `LineEntry` inside `Warning`.

First, refactored fixers to stop taking mutable `Warning` objects. Instead, they only talke line numbers where a fix is needed. Changed `fixes::run` to take a map of warning name to line numbers, instead of mutable `Warning`s. The caller `lib::fix` prepares this map by going over the warnings generated by the checker, and moves it over to `fixes::run`.

Updated `Warning` definition to store a borrowed reference. Update `Check` and its implementations to use explicit lifetimes for references passed to`run`.

NOTE: One side effect of this change is that warnings will be printed before backup path is printed.

With these changes, `cargo build` compiles fine. `cargo test` would require refactoring tests first, which would be simpler after #421 and #422 are merged.

#### ✔ Checklist:

### DIFF
--- a/src/common/warning.rs
+++ b/src/common/warning.rs
@@ -3,14 +3,18 @@ use std::fmt;
 use crate::common::*;
 
 #[derive(Clone, Debug, PartialEq)]
-pub struct Warning {
+pub struct Warning<'l> {
     pub check_name: String,
-    line: LineEntry,
+    line: &'l LineEntry,
     message: String,
 }
 
-impl Warning {
-    pub fn new(line: LineEntry, check_name: impl Into<String>, message: impl Into<String>) -> Self {
+impl<'l> Warning<'l> {
+    pub fn new(
+        line: &'l LineEntry,
+        check_name: impl Into<String>,
+        message: impl Into<String>,
+    ) -> Self {
         let check_name = check_name.into();
         let message = message.into();
         Self {
@@ -25,7 +29,7 @@ impl Warning {
     }
 }
 
-impl fmt::Display for Warning {
+impl<'l> fmt::Display for Warning<'l> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(
             f,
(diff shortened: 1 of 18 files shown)


mgrachev:
@atsmtat 👋 Thank you for your contribution ✌️

> NOTE: One side effect of this change is that warnings will be printed before backup path is printed.

It would be great to get the original behaviour back.

src/common/warning.rs
```diff
@@ -3,14 +3,18 @@ use std::fmt;
 use crate::common::*;
 
 #[derive(Clone, Debug, PartialEq)]
-pub struct Warning {
+pub struct Warning<'l> {
     pub check_name: String,
-    line: LineEntry,
+    line: &'l LineEntry,
     message: String,
 }
```
DDtKey, line 10: What do you think of a complete change in approach? 🤔 

As I see you have successfully abandoned `LineEntry` in `Fixers`, it seems to me that nothing prevents you from completely removing them from `Warnings`.

Where we need it, we will take the line by index, and at the same time we will completely solve the problem without lighttime or `Rc`.

Also, we can avoid the mentioned side effect with the output of warnings.
We will not need to modify them in the `HashMap`, we will be able to print them after the fixers work.

```suggestion
pub struct Warning {
    pub check_name: String,
    line_num: usize,
    message: String,
}
```

mgrachev: I like this idea 👍

mgrachev: @atsmtat What do you think about this?

mgrachev:
@atsmtat I'm sorry, but I'll close the PR.

When you are ready to finish this PR, please reopen it.

src/common/warning.rs
atsmtat: I like this idea too 👍 In fact, it crossed my mind before, but wasn't sure if you'll like it as it's a major change. Since we all like it, I gotta try it out 🙂

PR #450, opened 2021-11-20
feat: add support for multi-line single-quoted values

### LINKED ISSUE #442: Real multiple line problem
Hi,
But what about variables containing real newlines?
for example:

```
# .env.example
FOO=baz
MULTILINE_FOO='{
    "first": 1,
    "second": 1
  }'
BAR=baz
```
it is a real work example with `python-dotenv`
and dotenv-linter doesn't work with them:
```
dotenv-linter .env.example
Checking .env.example
.env.example:2 QuoteCharacter: The value has quote characters (', ")
.env.example:3 IncorrectDelimiter: The "first": 1, key has incorrect delimiter
.env.example:3 KeyWithoutValue: The "first": 1, key should be with a value or have an equal sign
.env.example:3 LeadingCharacter: Invalid leading character detected
.env.example:3 LowercaseKey: The "first": 1, key should be in uppercase
.env.example:3 UnorderedKey: The "first": 1, key should go before the FOO key
.env.example:4 IncorrectDelimiter: The "second": 1 key has incorrect delimiter
.env.example:4 KeyWithoutValue: The "second": 1 key should be with a value or have an equal sign
.env.example:4 LeadingCharacter: Invalid leading character detected
.env.example:4 LowercaseKey: The "second": 1 key should be in uppercase
.env.example:4 UnorderedKey: The "second": 1 key should go before the FOO key
.env.example:5 KeyWithoutValue: The }' key should be with a value or have an equal sign
.env.example:5 LeadingCharacter: Invalid leading character detected
.env.example:6 UnorderedKey: The BAR key should go before the FOO key

Found 14 problems
```

DDtKey:
Proof of concept for real multi-line value support #442.
~~It's temporary ugly solution, but it works.~~

Currently, I have added support of real multiline single-quoted values.
This implementation allows us to perform all checks on such values, since they are transformed into one `LineEntry`.

@dotenv-linter/core  I suggest to consider this PR and, if approved, add support of multiline  double-quoted values in another pull request (if needed at all) 🤔 

Partially closes #442

#### ✔ Checklist:

### DIFF
--- a/src/common.rs
+++ b/src/common.rs
@@ -23,6 +23,10 @@ pub fn remove_invalid_leading_chars(string: &str) -> &str {
     string.trim_start_matches(|c: char| !(c.is_alphabetic() || c == '_'))
 }
 
+pub fn is_escaped(prefix: &str) -> bool {
+    prefix.chars().rev().take_while(|ch| *ch == '\\').count() % 2 == 1
+}
+
 #[cfg(test)]
 pub(crate) mod tests {
     use super::*;
--- a/src/common.rs
+++ b/src/common.rs
@@ -97,4 +97,19 @@ pub(crate) mod tests {
         let string = "***FOO-BAR";
         assert_eq!("FOO-BAR", remove_invalid_leading_chars(string));
     }
+
+    #[test]
+    fn is_escaped_value_test() {
+        let escaped = "\\";
+        assert!(is_escaped(escaped));
+
+        let escaped = "\\\\\\";
+        assert!(is_escaped(escaped));
+
+        let non_escaped = "\\\\";
+        assert!(!is_escaped(non_escaped));
+
+        let random_string = "text without escaping";
+        assert!(!is_escaped(random_string));
+    }
 }
--- a/src/fixes.rs
+++ b/src/fixes.rs
@@ -19,7 +19,9 @@ trait Fix {
     fn fix_warnings(&self, warning_lines: &[usize], lines: &mut Vec<LineEntry>) -> Option<usize> {
         let mut count: usize = 0;
         for line_number in warning_lines {
-            let line = lines.get_mut(line_number - 1)?;
+            let line = lines
+                .iter_mut()
+                .find(|entry| entry.number == *line_number)?;
             if self.fix_line(line).is_some() {
                 count += 1;
             }
@@ -143,7 +145,7 @@ mod tests {
             ),
         ];
 
-        assert_eq!(0, run(&mut warnings, &mut lines, &[]));
+        assert_eq!(2, run(&mut warnings, &mut lines, &[]));
     }
 
     #[test]
--- a/src/fixes/unordered_key.rs
+++ b/src/fixes/unordered_key.rs
@@ -68,7 +68,7 @@ impl Fix for UnorderedKeyFixer {
                 }
 
                 if line.is_empty()
-                    || lines.len() == line.number // Is this the last line?
+                    || lines.len() == i + 1 // Is this the last line?
                     || is_control_comment
                     || has_substitution_variables
                 {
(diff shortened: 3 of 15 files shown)


tests/fixes/fixtures/complicated.env
```diff
@@ -70,4 +70,16 @@ VAR_4=1234
 VAR_3=1234
 VAR_2=1234
 VAR_1=1234
-ENV24='key56'
\ No newline at end of file
+ENV24='key56'
+# dotenv-linter:on QuoteCharacter
+
+# multiline values & ordering:
```
DDtKey, line 76: This block shows that it works as expected: values will be trimmed and reordered

src/common.rs
```diff
@@ -23,6 +23,10 @@ pub fn remove_invalid_leading_chars(string: &str) -> &str {
     string.trim_start_matches(|c: char| !(c.is_alphabetic() || c == '_'))
 }
 
+pub fn is_escaped(prefix: &str) -> bool {
+    prefix.chars().rev().take_while(|ch| *ch == '\\').count() % 2 == 1
```
mgrachev, line 27: Could you add tests for this function?

src/fixes.rs
```diff
@@ -19,7 +19,9 @@ trait Fix {
     fn fix_warnings(&self, warning_lines: &[usize], lines: &mut Vec<LineEntry>) -> Option<usize> {
         let mut count: usize = 0;
         for line_number in warning_lines {
-            let line = lines.get_mut(line_number - 1)?;
+            let line = lines
+                .iter_mut()
+                .find(|entry| entry.number == *line_number)?;
```
mgrachev, line 24: Why did you decide to rewrite this code?

src/fixes.rs
```diff
@@ -143,7 +145,7 @@ mod tests {
             ),
         ];
 
-        assert_eq!(0, run(&mut warnings, &mut lines, &[]));
+        assert_eq!(2, run(&mut warnings, &mut lines, &[]));
```
mgrachev, line 148: Could you tell me more about this change?
Has this test been wrong before?

src/fixes/unordered_key.rs
```diff
@@ -68,7 +68,7 @@ impl Fix for UnorderedKeyFixer {
                 }
 
                 if line.is_empty()
-                    || lines.len() == line.number // Is this the last line?
+                    || lines.len() == i + 1 // Is this the last line?
```
mgrachev, line 71: Could you tell me more about this change?
What's wrong with `line.number`?

src/fixes.rs
DDtKey: Because line number != index.
Multiline values contain multiple line numbers in one.

And this led to a strong dependence of the order of fixes (e.g, ordering)

src/common.rs
DDtKey: Sure!

src/fixes.rs
DDtKey: Actually, this test seemed a bit strange. But yeah, previously, warnings were not fixed if the order of the lines was wrong.  

I think this behavior makes sense now.  Because, the warnings are attached to the original lines, this should not depend on `LineEntry` ordering

In fact, now UnorderedKeyFixer should not break the behavior of others.

src/fixes/unordered_key.rs
DDtKey: Well, for the same reason as [here](https://github.com/dotenv-linter/dotenv-linter/pull/450#discussion_r757765209)

src/common.rs
DDtKey: Done

mgrachev (approved):

mgrachev:
@DDtKey Thanks a lot 👍

PR #453, opened 2021-12-12
feat: add support for double quoted multi-line values

### LINKED ISSUE #442: Real multiple line problem

DDtKey:
Continuing to #450

Closes #442

It seems that we should update documentation. But tech-functionality is ready 🙂 

#### ✔ Checklist:

- [X] Commit messages have been written in [Conventional Commits](https://www.conventionalcommits.org) format;

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,6 +5,7 @@ use std::str::FromStr;
 use crate::common::*;
 
 pub use checks::available_check_names;
+use common::quote_type::QuoteType;
 mod checks;
 mod common;
 mod fixes;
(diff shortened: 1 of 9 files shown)


src/lib.rs
```diff
@@ -269,34 +270,39 @@ fn reduce_multiline_entries(lines: &mut Vec<LineEntry>) {
...
-            if let Some(idx) = entry.raw_string.find('\'') {
-                if !is_escaped(&entry.raw_string[..idx]) {
-                    multiline_ranges.push((start, entry.number));
+            if let Some(quote_char) = quote_char {
+                if let Some(idx) = entry.raw_string.find(quote_char) {
+                    if !is_escaped(&entry.raw_string[..idx]) {
+                        multiline_ranges.push((start, entry.number));
+                        start_number = None;
+                    }
+                } else if entry.get_value().is_some() {
```
DDtKey, line 284: This was done in a previous commit (#450)
However, I am still afraid that it could have been done in another way.

I did this with backward compatibility to check for values starting with a quotation mark (not multi-line, see `QuoteCharacterChecker`). But what if multiline can contain char `=`?

We have to find the right way to distinguish between a valid `env` string from a multiline string with the `=` char

src/lib.rs
```diff
@@ -269,34 +270,39 @@ fn reduce_multiline_entries(lines: &mut Vec<LineEntry>) {
...
+                    // if next line correct env line - then previous start-line incorrect multi-value
```
DDtKey, line 285: This was done in a previous commit (#450)
However, I am still afraid that it could have been done in another way.

I did this with backward compatibility to check for values starting with a quotation mark (not multi-line, see `QuoteCharacterChecker`). But what if multiline can contain char `=`?

We have to find the right way to distinguish between a valid `env` string from a multiline string with the `=` char

This can be left to change later, but should not be forgotten 🤔

mgrachev: Yes, you're right. Please open an issue so as not to forget.

mgrachev (approved):

PR #454, opened 2021-12-30
Add check for new version

mgrachev:
#### ✔ Checklist:

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,11 +1,14 @@
+use crate::common::*;
+use crate::quote_type::QuoteType;
+use colored::*;
 use std::collections::{BTreeMap, HashSet};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
-
-use crate::common::*;
+use std::time::Duration;
+use update_informer::registry::Crates;
 
 pub use checks::available_check_names;
-use common::quote_type::QuoteType;
+
 mod checks;
 mod common;
 mod fixes;
@@ -21,6 +24,8 @@ pub fn check(args: &clap::ArgMatches, current_dir: &Path) -> Result<usize> {
 
     if lines_map.is_empty() {
         output.print_nothing_to_check();
+        print_new_version_if_available(args);
+
         return Ok(0);
     }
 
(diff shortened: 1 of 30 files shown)


Dockerfile
```diff
@@ -1,9 +1,6 @@
-FROM rust:latest as builder
-WORKDIR /usr/src
-RUN rustup target add x86_64-unknown-linux-musl
+FROM messense/rust-musl-cross:x86_64-musl as builder
```
DDtKey, line 1: It seems that this file should be rebased after #456

mgrachev: Already has done 🙂

src/lib.rs
```diff
@@ -306,3 +311,44 @@ fn is_multiline_start(val: &str) -> Option<QuoteType> {
...
+    #[cfg(not(feature = "stub_check_version"))]
+    let current_version = env!("CARGO_PKG_VERSION");
+
+    #[cfg(feature = "stub_check_version")]
+    let current_version = "3.0.0";
+
+    let pkg_name = env!("CARGO_PKG_NAME");
+    let interval = Duration::from_secs(60 * 60 * 24);
+
+    #[cfg(not(feature = "stub_check_version"))]
```
DDtKey, line 330: Honestly, call different functions depends on feature - looks a bit weird. I see high coupling in this external functions.
It seems like this only needed for test purposes. 

I'd prefer to have api with different instances/structs of `UpdateInformer` - but with single interface/trait
Like this:

```rust
#[cfg(not(feature = "stub_check_version"))]
let update_informer = UpdateInformer::new(Crates, interval); // it seems that `pkg_name` and `version` could be moved to constructor (fn new), because `env!` macro will be expanded on compile time

#[cfg(feature = "stub_check_version")]
let update_informer = StubUpdateInformer::new(Crates, interval, "3.1.1");

let result = update_informer.check_version(); // check_version - fn from `Trait` implemented in different ways for Stub/non-stub
```

Is it possible to make similar changes in the dependency part? 
If you disagree, it's okay, let's leave it this way

mgrachev: Thank you for your feedback!
I like your option, but I'll do it a bit later.

PR #625, opened 2023-01-14
feat: Allow setting not_check_updates flag via env var

### LINKED ISSUE #593: Add environment variable for --not-check-updates
In certain environments, it may not be feasible to control all invocations of dotenv-linter. However the environment can typically still be controlled (and usually more easily even if invocations can be controlled). It would be nice if there was a way to set this with an environment variable.

jackyef:
#### ✔ Checklist:

Closes #593

This is my very first Rust PR, any feedback and pointers are very much appreciated!

### DIFF
--- a/dotenv-linter/src/cli/mod.rs
+++ b/dotenv-linter/src/cli/mod.rs
@@ -154,6 +154,8 @@ fn not_check_updates_flag() -> Arg {
     Arg::new("not-check-updates")
         .long("not-check-updates")
         .help("Doesn't check for updates")
+        .value_parser(clap::builder::BoolishValueParser::new())
+        .env("DOTENV_LINTER_NOT_CHECK_UPDATES")
         .action(ArgAction::SetTrue)
 }
 
--- a/dotenv-linter/tests/common/test_dir.rs
+++ b/dotenv-linter/tests/common/test_dir.rs
@@ -29,10 +29,10 @@ impl TestDir {
 
     // Only used in tests
     #[allow(dead_code)]
-    pub fn with_envs(envs: Option<HashMap<String, String>>) -> Self {
+    pub fn with_envs(envs: HashMap<String, String>) -> Self {
         Self {
             current_dir: tempdir().expect("create testdir with envs"),
-            envs: envs.clone(),
+            envs: Some(envs.clone()),
         }
     }
 
@@ -89,7 +89,7 @@ impl TestDir {
         T: Into<String>,
     {
         let expected_output = expected_output.into();
-        let mut cmd = Self::init_cmd(self.envs.clone());
+        let mut cmd = self.init_cmd();
         let canonical_current_dir = canonicalize(&self.current_dir).expect("canonical current dir");
         cmd.current_dir(&canonical_current_dir)
             .args(args)
@@ -111,7 +111,7 @@ impl TestDir {
         T: Into<String>,
     {
         let expected_output = expected_output.into();
-        let mut cmd = Self::init_cmd(self.envs.clone());
+        let mut cmd = self.init_cmd();
         let canonical_current_dir = canonicalize(&self.current_dir).expect("canonical current dir");
         cmd.current_dir(&canonical_current_dir)
             .args(args)
@@ -131,7 +131,7 @@ impl TestDir {
         T: Into<String>,
     {
         let expected_output = expected_output.into();
-        let mut cmd = Self::init_cmd(self.envs.clone());
+        let mut cmd = self.init_cmd();
         let canonical_current_dir = canonicalize(&self.current_dir).expect("canonical current dir");
         cmd.current_dir(&canonical_current_dir)
             .args(&["fix", "--no-backup"])
@@ -144,7 +144,7 @@ impl TestDir {
     ///
     /// This method does NOT remove TestDir when finished
     pub fn test_command_fix_success_without_output(&self) {
-        let mut cmd = Self::init_cmd(self.envs.clone());
+        let mut cmd = self.init_cmd();
         let canonical_current_dir = canonicalize(&self.current_dir).expect("canonical current dir");
         cmd.current_dir(&canonical_current_dir)
             .args(&["fix", "--no-backup"])
@@ -163,7 +163,7 @@ impl TestDir {
         T: Into<String>,
     {
         let expected_output = expected_output.into();
-        let mut cmd = Self::init_cmd(self.envs.clone());
+        let mut cmd = self.init_cmd();
         let canonical_current_dir = canonicalize(&self.current_dir).expect("canonical current dir");
         cmd.current_dir(&canonical_current_dir)
             .args(&["fix", "--no-backup"])
@@ -182,7 +182,7 @@ impl TestDir {
         I: IntoIterator<Item = S>,
         S: AsRef<OsStr>,
     {
-        let mut cmd = Self::init_cmd(self.envs.clone());
+        let mut cmd = self.init_cmd();
         let canonical_current_dir = canonicalize(&self.current_dir).expect("canonical current dir");
         cmd.current_dir(&canonical_current_dir)
             .args(args)
@@ -199,7 +199,7 @@ impl TestDir {
         I: IntoIterator<Item = S>,
         S: AsRef<OsStr>,
     {
-        let mut cmd = Self::init_cmd(self.envs.clone());
+        let mut cmd = self.init_cmd();
         let canonical_current_dir = canonicalize(&self.current_dir).expect("canonical current dir");
         String::from(
             from_utf8(
@@ -215,10 +215,10 @@ impl TestDir {
         )
     }
 
-    fn init_cmd(envs: Option<HashMap<String, String>>) -> Command {
+    fn init_cmd(&self) -> Command {
         let mut cmd = Command::cargo_bin(env!("CARGO_PKG_NAME")).expect("command from binary name");
 
-        cmd.envs(envs.unwrap_or_default());
+        cmd.envs(&self.envs.clone().unwrap_or_default());
 
         cmd
     }
--- a/dotenv-linter/tests/common/test_dir.rs
+++ b/dotenv-linter/tests/common/test_dir.rs
@@ -1,4 +1,5 @@
 use assert_cmd::Command;
+use std::collections::HashMap;
 use std::{borrow::Cow, ffi::OsStr};
 use tempfile::{tempdir, tempdir_in, TempDir};
 
@@ -14,6 +15,7 @@ use std::str::from_utf8;
 /// Use to test commands in temporary directories
 pub struct TestDir {
     current_dir: TempDir,
+    envs: Option<HashMap<String, String>>,
 }
 
 impl TestDir {
@@ -21,6 +23,16 @@ impl TestDir {
     pub fn new() -> Self {
         Self {
             current_dir: tempdir().expect("create testdir"),
+            envs: None,
+        }
+    }
+
+    // Only used in tests
+    #[allow(dead_code)]
+    pub fn with_envs(envs: Option<HashMap<String, String>>) -> Self {
+        Self {
+            current_dir: tempdir().expect("create testdir with envs"),
+            envs: envs.clone(),
         }
     }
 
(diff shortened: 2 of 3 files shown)


dotenv-linter/src/cli/mod.rs
```diff
@@ -154,6 +154,8 @@ fn not_check_updates_flag() -> Arg {
     Arg::new("not-check-updates")
         .long("not-check-updates")
         .help("Doesn't check for updates")
+        .value_parser(clap::builder::BoolValueParser::new())
```
DDtKey, line 157: I'd like to use this one: [BoolishValueParser](https://docs.rs/clap/latest/clap/builder/struct.BoolishValueParser.html)

DDtKey (commented):
Thank you for the contribution! 👍

dotenv-linter/tests/flags/check_updates.rs
```diff
@@ -26,3 +27,17 @@ fn print_new_version_if_nothing_to_check() {
     let args: &[&str; 0] = &[];
     test_dir.test_command_success_with_args(args, expected_output);
 }
+
+#[test]
+fn do_not_print_new_version() {
+    env::set_var("DOTENV_LINTER_NOT_CHECK_UPDATES", "true");
```
DDtKey, line 33: It also would be nice to pass this env to `cmd` ([doc](https://doc.rust-lang.org/std/process/struct.Command.html#method.env)), otherwise there is a chance of parallel run with other tests with enabled variable. So it could silently affect other tests.

Take a look at `TestDir`, we need to extend it to allow passing `ENV` variables for commands.

jackyef: I added a commit for this here: https://github.com/dotenv-linter/dotenv-linter/pull/625/commits/6ff87026df99705446705fc338333d81e27a81fe

I am not too sure if I am doing it the right way though. It feels like I am making too many changes, please let me know if there's an obvious way I am missing! 🙇

dotenv-linter/src/cli/mod.rs
jackyef: Updated here: https://github.com/dotenv-linter/dotenv-linter/pull/625/commits/005622de4a6c0a0df564c092a07baf4e3e395404#diff-345b89c66e65c7e3c0f07e0d58c74a65f44215fcdf6e24dfd32a11f89f7209e5R157

dotenv-linter/tests/common/test_dir.rs
```diff
@@ -14,20 +15,23 @@ use std::str::from_utf8;
 /// Use to test commands in temporary directories
 pub struct TestDir {
     current_dir: TempDir,
+    envs: Option<HashMap<String, String>>,
 }
 
 impl TestDir {
     /// Generate a new TestDir
-    pub fn new() -> Self {
+    pub fn new(envs: Option<HashMap<String, String>>) -> Self {
```
DDtKey, line 23: You can don't change `new`, just encapsulate `None` there. And add one more constructor `with_envs`.

Second option is pass ENV variables directly to methods as optional argument, but it would involve all usages. So it could be also two- definition per method (to avoid changes in other places)

dotenv-linter/tests/flags/check_updates.rs
DDtKey: Here the possible options are: https://github.com/dotenv-linter/dotenv-linter/pull/625#discussion_r1070291629

dotenv-linter/tests/common/test_dir.rs
jackyef: Thank you for the pointers, they were really helpful! Updated here: https://github.com/dotenv-linter/dotenv-linter/pull/625/commits/7d85e99ffc5c79e4c52eb386e7875cf5d2aaa6a9

dotenv-linter/tests/common/test_dir.rs
```diff
@@ -14,20 +15,32 @@ use std::str::from_utf8;
...
     pub fn new() -> Self {
         Self {
             current_dir: tempdir().expect("create testdir"),
+            envs: None,
+        }
+    }
+
+    // Only used in tests
+    #[allow(dead_code)]
+    pub fn with_envs(envs: Option<HashMap<String, String>>) -> Self {
```
DDtKey, line 32: There is no need to keep this method with optional param, it's supposed to be always with envs, otherwise `new` exists

dotenv-linter/tests/common/test_dir.rs
```diff
@@ -14,20 +15,32 @@ use std::str::from_utf8;
...
     /// Generate a new TestDir
     pub fn new() -> Self {
         Self {
             current_dir: tempdir().expect("create testdir"),
+            envs: None,
+        }
+    }
+
+    // Only used in tests
+    #[allow(dead_code)]
```
DDtKey, line 31: Do we really need it? 🤔 `new` also used only in tests AFAIR. Usage exists, so it should be a dead code

dotenv-linter/tests/common/test_dir.rs
```diff
@@ -118,7 +131,7 @@ impl TestDir {
         T: Into<String>,
     {
         let expected_output = expected_output.into();
-        let mut cmd = Self::init_cmd();
+        let mut cmd = Self::init_cmd(self.envs.clone());
```
DDtKey, line 134: Btw, looks like you can avoid extra cloning in two ways:
1. Inplace env command, to make it like this:
```
let mut cmd = Self::init_cmd();
cmd.envs(&self.envs);
````
Looks like it should work.

But second option could be even better, declare function `init_cmd` as method. So it will be: `self.init_cmd();` and will pass envs as reference to cmd

**Review by DDtKey — COMMENTED** [review]
Thanks for addressing all comments, but if would like to improve it a bit more, there are a several tips/suggestions

File: dotenv-linter/tests/common/test_dir.rs (hunk 5 above)
Comment (jackyef) [reply dotenv-linter/tests/common/test_dir.rs:31]: I think it may have something to do with the test being part of a specific features.

Running `cargo t`, a warning will be logged about this:
```
cargo t
   Compiling dotenv-linter v3.3.0 (/home/jackyef/Projects/dotenv-linter/dotenv-linter)
warning: associated function `with_envs` is never used
  --> dotenv-linter/tests/common/test_dir.rs:30:12
   |
30 |     pub fn with_envs(envs: HashMap<String, String>) -> Self {
   |            ^^^^^^^^^
   |
   = note: `#[warn(dead_code)]` on by default

^[[Awarning: `dotenv-linter` (test "cli") generated 1 warning
```

The warning won't appear when running `cargo t --all-features` though! Is it preferable to remove this anyway?

File: dotenv-linter/tests/common/test_dir.rs (hunk 4 above)
Comment (jackyef) [reply dotenv-linter/tests/common/test_dir.rs]: Updated here! https://github.com/dotenv-linter/dotenv-linter/pull/625/commits/c64b3667d04861528bf4a93e506ffca117fb4e07

File: dotenv-linter/tests/common/test_dir.rs (hunk 6 above)
Comment (jackyef) [reply dotenv-linter/tests/common/test_dir.rs]: Thanks! Updated here: https://github.com/dotenv-linter/dotenv-linter/pull/625/commits/c64b3667d04861528bf4a93e506ffca117fb4e07

Comment (jackyef) [reply dotenv-linter/tests/common/test_dir.rs]: Appreciate all the suggestions by the way, really helpful for me to get myself more familiar with the language! :)

File: dotenv-linter/tests/common/test_dir.rs [hunk 7]
```diff
@@ -202,7 +215,11 @@ impl TestDir {
         )
     }
 
-    fn init_cmd() -> Command {
-        Command::cargo_bin(env!("CARGO_PKG_NAME")).expect("command from binary name")
+    fn init_cmd(&self) -> Command {
+        let mut cmd = Command::cargo_bin(env!("CARGO_PKG_NAME")).expect("command from binary name");
+
+        cmd.envs(&self.envs.clone().unwrap_or_default());
```
DDtKey, line 221: Doesn't it work with reference?
Something like that `self.envs.as_ref().unwrap_or(&[])`

dotenv-linter/tests/common/test_dir.rs
DDtKey: Usually we use `all-features` ([CI workflow](https://github.com/dotenv-linter/dotenv-linter/blob/master/.github/workflows/ci.yml)), however you can leave it as it is. I don't mind

dotenv-linter/tests/common/test_dir.rs
DDtKey: And take a look at CI warnings please.

jackyef: Ah, I wasn't familiar with `as_ref()`, this should do it. https://github.com/dotenv-linter/dotenv-linter/pull/625/commits/9eeead1552b1a18bc839726f37ba7ab5f2305f36
Also ran the formatter as well 👌

DDtKey (approved):
Thank you @jackyef for implementing that!

Commilint is failing, however I'd suggest to disable it for commits inside PR, we always use squash strategy. So only PR name is important and commits inside `main` branch. 
@dotenv-linter/core WDYT? I believe we can enable something like that for PRs:
```
      - uses: CondeNast/conventional-pull-request-action@v0.2.0
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        with:
          ignoreCommits: "true"
```

dotenv-linter/tests/common/test_dir.rs
```diff
@@ -14,20 +15,32 @@ use std::str::from_utf8;
 /// Use to test commands in temporary directories
 pub struct TestDir {
     current_dir: TempDir,
+    envs: Option<HashMap<String, String>>,
```
mgrachev, line 18:
```suggestion
    envs: HashMap<String, String>,
```
You can use an empty hash map instead of `Option<HashMap>`.

dotenv-linter/tests/common/test_dir.rs
```diff
@@ -14,20 +15,32 @@ use std::str::from_utf8;
...
 }
 
 impl TestDir {
     /// Generate a new TestDir
     pub fn new() -> Self {
         Self {
             current_dir: tempdir().expect("create testdir"),
+            envs: None,
```
mgrachev, line 26:
```suggestion
            envs: Default::default(),
```
It's alternative to use `HashMap::new()`.

dotenv-linter/tests/common/test_dir.rs
```diff
@@ -14,20 +15,32 @@ use std::str::from_utf8;
...
+        }
+    }
+
+    // Only used in tests
+    #[allow(dead_code)]
+    pub fn with_envs(envs: HashMap<String, String>) -> Self {
+        Self {
+            current_dir: tempdir().expect("create testdir with envs"),
+            envs: Some(envs.clone()),
```
mgrachev, line 35:
```suggestion
            envs,
```
You don't need to clone this argument.

dotenv-linter/tests/common/test_dir.rs
```diff
@@ -202,7 +215,11 @@ impl TestDir {
         )
     }
 
-    fn init_cmd() -> Command {
-        Command::cargo_bin(env!("CARGO_PKG_NAME")).expect("command from binary name")
+    fn init_cmd(&self) -> Command {
+        let mut cmd = Command::cargo_bin(env!("CARGO_PKG_NAME")).expect("command from binary name");
+
+        cmd.envs(self.envs.as_ref().unwrap_or(&HashMap::new()));
```
mgrachev, line 221:
```suggestion
        cmd.envs(&self.envs);
```

mgrachev (commented):
I've left some suggestions too 🙂

mgrachev:
> Commilint is failing, however I'd suggest to disable it for commits inside PR, we always use squash strategy. So only PR name is important and commits inside `main` branch. @dotenv-linter/core WDYT? I believe we can enable something like that for PRs:
> 
> ```
>       - uses: CondeNast/conventional-pull-request-action@v0.2.0
>         env:
>           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
>         with:
>           ignoreCommits: "true"
> ```

I don't mind. Can you open a PR with these changes? 🙂

DDtKey (approved):

mgrachev (approved):

mgrachev:
@jackyef Thank you for your help 🔥

jackyef:
Thank you both for your guidance! 🙌

Issue #673, opened 2023-05-31
Incorrect fix for JSON string env vars


ducminh-phan:
Currently, we are using [pydantic](https://docs.pydantic.dev/latest/usage/settings/#parsing-environment-variable-values) to read settings from environment variables. For complex types, `list[str]` for example, the values need to be JSON-encoded strings. For example,

```
FOO='["a","b"]'
```

With this dotenv file, `dotenv-linter fix` changes the file to

```
FOO=[a,b]
```

The value now is not a valid JSON string, therefore, `pydantic` could not parse the environment variable.

`FOO='["a", "b"]'` works, since it has a space in the value, but there's no workaround for `FOO='["a"]'`.

mgrachev:
@ducminh-phan Hi 👋 Thank you for your feedback!

[We decided](https://github.com/dotenv-linter/dotenv-linter/pull/678#discussion_r1224621044) not to support JSON arrays in out tool.

> FOO='["a", "b"]' works, since it has a space in the value, but there's no workaround for FOO='["a"]'.

As a workaround, use the control comment to disable this check:
```
# dotenv-linter:off QuoteCharacter
FOO='["a"]'
# dotenv-linter:on QuoteCharacter
```

ducminh-phan:
@mgrachev Thanks for getting back. I understand the decision to not support JSON string env vars. `# dotenv-linter:off QuoteCharacter` is good enough for my usecase 👍

PR #678, opened 2023-06-08
fix: don't remove quotes for json str arrays

motherofdaemons:
I was looking for open source to contribute to and found this #673 was open and thought it would be a good first issue. I believe that this should catch all single line JSON str arrays.

#### ✔ Checklist:


dotenv-linter/src/checks/quote_character.rs
```diff
@@ -31,6 +31,10 @@ impl Check for QuoteCharacterChecker<'_> {
             return None;
         }
 
+        if val.contains('\'') && val.contains('\"') && val.contains('[') && val.contains(']') {
```
mgrachev, line 34: Please add a comment on which case we are handling here.

dotenv-linter/src/checks/quote_character.rs
```diff
@@ -31,6 +31,10 @@ impl Check for QuoteCharacterChecker<'_> {
             return None;
         }
```
mgrachev, line 33: Please add a comment on which case we are handling here:
```suggestion
// Don't remove quotes for arrays of JSON strings, e.g.: '["a"]'
```

motherofdaemons: I have added the comment! Thanks for the suggestion!

dotenv-linter/src/checks/quote_character.rs
```diff
@@ -31,6 +31,11 @@ impl Check for QuoteCharacterChecker<'_> {
...
 
+        // Don't remove quotes for arrays of JSON strings, e.g.: '["a"]'
+        if val.contains('\'') && val.contains('\"') && val.contains('[') && val.contains(']') {
```
DDtKey, line 35: Sorry, by I'm not sure of this logic to be honest  🤔 

It doesn't guarantee that line is correct JSON, it might be something like `'...]"['` and this check will ignore this?
Is there a similar behavior in other env linters or libs? It seems too specific to cover JSON arrays separately

If we would like to support such cases - it should be somehow parsed to [serde_json::Value](https://docs.rs/serde_json/latest/serde_json/value/enum.Value.html) for example 🤔 

However, it might be betterto disable  `QuoteCharacterChecker` for specific lines  in such cases instead of supporting all possible cases.
I'm even thinking if it should be optional checker, because it's not really breaks ENV format in any way and users might want to wrap values into `'`/`'"` due to different reasons.

@dotenv-linter/core WDYT?

DDtKey (changes requested):

dotenv-linter/src/checks/quote_character.rs
mstruebing: My naive view is that we should disable the `QuoteCharacterChecker` in such cases as we otherwise would run into a situation where we would need to support more and more special cases for other formats.

mgrachev: > However, it might be better to disable QuoteCharacterChecker for specific lines in such cases instead of supporting all possible cases on linter side.

I agree. Do I understand correctly that the user himself should disable this check using the control comments?
```
# dotenv-linter:off QuoteCharacter
FOO='["a","b"]'
```

DDtKey: Yes, I meant this way

mgrachev:
@motherofdaemons Sorry, we decided not to support this case in our tool.

PR #761, opened 2024-03-05
feat: new schema based checker

pm100:
I do not expect this to be accepted as is. I am sure there will be feedback

There are a couple of big changed to the structure of the code

First I have unified the 3 different Options structs (Cli, Fix, Compare) into one CliOptions. Why, becuase its seems useful to pass the options to the checkers, rather than pull out one field (skips). But fix accepts options but calls check and fix, so either I had to synthesize a checkoptions from the fixoptions, or just make it one options type.

Next, I have added a call to the Check trait that allows them to be called after all lines have been processed. Useful for detecting missing things. Check::run loops over the checker at the end calling this new entry point.

Features
- specify what keys are allowed
- specify the type, String, Number, Url, Email, Boolean
- specify required or not
- specify regex for string
- specify if other keys other than those listed are allowed.

sample schema

```
{
    "version": "1.0.0",
    "allow_other_keys": true,
    "entries": [
        {
            "key": "SECRET",
            "required": true,
            "type": "String",
            "regex": "^[a-z]*$"
        },
        {
            "key": "PORT",
            "atype": "Number"
        },
        {
            "key": "URL",
            "type": "Url"
        },
        {
            "key": "EMAIL",
            "type": "Email"
        }
    ]
}
```

to use include -S | --schema on command line

if schema not found or is invalid then error message is printed and app exits

Not done yet
- allowed set of values for a string or number
- allowed range for numeric

Full set of tests included

### DIFF
--- a/dotenv-linter/src/checks.rs
+++ b/dotenv-linter/src/checks.rs
@@ -1,3 +1,4 @@
+use crate::cli::options::CliOptions;
 use crate::common::comment::Comment;
 use crate::common::{LintKind, Warning};
 use dotenv_lookup::LineEntry;
@@ -10,6 +11,7 @@ mod key_without_value;
 mod leading_character;
 mod lowercase_key;
 mod quote_character;
+mod schema_validator;
 mod space_character;
 mod substitution_key;
 mod trailing_whitespace;
@@ -23,10 +25,13 @@ pub trait Check {
     fn skip_comments(&self) -> bool {
         true
     }
+    fn end(&mut self) -> Vec<Warning> {
+        vec![]
+    }
 }
 
 // Checklist for checks which needs to know of only a single line
-fn checklist() -> Vec<Box<dyn Check>> {
+fn checklist<'a>(opts: &'a CliOptions) -> Vec<Box<dyn Check + 'a>> {
     vec![
         Box::<duplicated_key::DuplicatedKeyChecker>::default(),
         Box::<ending_blank_line::EndingBlankLineChecker>::default(),
@@ -41,11 +46,13 @@ fn checklist() -> Vec<Box<dyn Check>> {
         Box::<trailing_whitespace::TrailingWhitespaceChecker>::default(),
         Box::<unordered_key::UnorderedKeyChecker>::default(),
         Box::<value_without_quotes::ValueWithoutQuotesChecker>::default(),
+        Box::new(schema_validator::SchemaChecker::new(opts)),
     ]
 }
 
-pub fn run(lines: &[LineEntry], skip_checks: &[LintKind]) -> Vec<Warning> {
-    let mut checks = checklist();
+pub fn run(lines: &[LineEntry], opts: &CliOptions) -> Vec<Warning> {
+    let mut checks = checklist(opts);
+    let skip_checks = &opts.skip;
 
     // Skip checks with the --skip argument (globally)
     checks.retain(|c| !skip_checks.contains(&c.name()));
@@ -81,6 +88,11 @@ pub fn run(lines: &[LineEntry], skip_checks: &[LintKind]) -> Vec<Warning> {
         }
     }
 
+    for ch in &mut checks {
+        let end_warns = ch.end();
+        warnings.extend(end_warns);
+    }
+
     warnings
 }
 
@@ -93,18 +105,16 @@ mod tests {
     fn run_with_empty_vec_test() {
         let empty: Vec<LineEntry> = Vec::new();
         let expected: Vec<Warning> = Vec::new();
-        let skip_checks: Vec<LintKind> = Vec::new();
-
-        assert_eq!(expected, run(&empty, &skip_checks));
+        let opt = CliOptions::default();
+        assert_eq!(expected, run(&empty, &opt));
     }
 
     #[test]
     fn run_with_empty_line_test() {
         let lines: Vec<LineEntry> = vec![blank_line_entry(1, 1)];
         let expected: Vec<Warning> = Vec::new();
-        let skip_checks: Vec<LintKind> = Vec::new();
-
-        assert_eq!(expected, run(&lines, &skip_checks));
+        let opt = CliOptions::default();
+        assert_eq!(expected, run(&lines, &opt));
     }
 
     #[test]
@@ -114,18 +124,16 @@ mod tests {
             blank_line_entry(2, 2),
         ];
         let expected: Vec<Warning> = Vec::new();
-        let skip_checks: Vec<LintKind> = Vec::new();
-
-        assert_eq!(expected, run(&lines, &skip_checks));
+        let opt = CliOptions::default();
+        assert_eq!(expected, run(&lines, &opt));
     }
 
     #[test]
     fn run_with_valid_line_test() {
         let lines: Vec<LineEntry> = vec![line_entry(1, 2, "FOO=BAR"), blank_line_entry(2, 2)];
         let expected: Vec<Warning> = Vec::new();
-        let skip_checks: Vec<LintKind> = Vec::new();
-
-        assert_eq!(expected, run(&lines, &skip_checks));
+        let opt = CliOptions::default();
+        assert_eq!(expected, run(&lines, &opt));
     }
 
     #[test]
@@ -138,9 +146,8 @@ mod tests {
         );
         let lines: Vec<LineEntry> = vec![line, blank_line_entry(2, 2)];
         let expected: Vec<Warning> = vec![warning];
-        let skip_checks: Vec<LintKind> = Vec::new();
-
-        assert_eq!(expected, run(&lines, &skip_checks));
+        let opt = CliOptions::default();
+        assert_eq!(expected, run(&lines, &opt));
     }
 
     #[test]
@@ -153,9 +160,8 @@ mod tests {
         );
         let lines: Vec<LineEntry> = vec![line];
         let expected: Vec<Warning> = vec![warning];
-        let skip_checks: Vec<LintKind> = Vec::new();
-
-        assert_eq!(expected, run(&lines, &skip_checks));
+        let opt = CliOptions::default();
+        assert_eq!(expected, run(&lines, &opt));
     }
 
     #[test]
@@ -169,9 +175,10 @@ mod tests {
         );
         let lines: Vec<LineEntry> = vec![line1, line2, blank_line_entry(3, 3)];
         let expected: Vec<Warning> = vec![warning];
-        let skip_checks: Vec<LintKind> = vec![LintKind::KeyWithoutValue, LintKind::UnorderedKey];
+        let mut opt = CliOptions::default();
+        opt.skip = vec![LintKind::KeyWithoutValue, LintKind::UnorderedKey];
 
-        assert_eq!(expected, run(&lines, &skip_checks));
+        assert_eq!(expected, run(&lines, &opt));
     }
 
     #[test]
@@ -179,9 +186,10 @@ mod tests {
         let line = line_entry(1, 1, "FOO");
         let lines: Vec<LineEntry> = vec![line];
         let expected: Vec<Warning> = Vec::new();
-        let skip_checks: Vec<LintKind> = vec![LintKind::KeyWithoutValue, LintKind::EndingBlankLine];
+        let mut opt = CliOptions::default();
+        opt.skip = vec![LintKind::KeyWithoutValue, LintKind::EndingBlankLine];
 
-        assert_eq!(expected, run(&lines, &skip_checks));
+        assert_eq!(expected, run(&lines, &opt));
     }
 
     #[test]
@@ -196,9 +204,10 @@ mod tests {
         );
         let lines: Vec<LineEntry> = vec![line1, line2, line3, blank_line_entry(4, 4)];
         let expected: Vec<Warning> = vec![warning];
-        let skip_checks: Vec<LintKind> = vec![LintKind::UnorderedKey];
+        let mut opt = CliOptions::default();
+        opt.skip = vec![LintKind::UnorderedKey];
 
-        assert_eq!(expected, run(&lines, &skip_checks));
+        assert_eq!(expected, run(&lines, &opt));
     }
 
     #[test]
@@ -213,9 +222,9 @@ mod tests {
         );
         let lines: Vec<LineEntry> = vec![line1, line2, line3, blank_line_entry(4, 4)];
         let expected: Vec<Warning> = vec![warning];
-        let skip_checks: Vec<LintKind> = vec![LintKind::KeyWithoutValue, LintKind::UnorderedKey];
-
-        assert_eq!(expected, run(&lines, &skip_checks));
+        let mut opt = CliOptions::default();
+        opt.skip = vec![LintKind::KeyWithoutValue, LintKind::UnorderedKey];
+        assert_eq!(expected, run(&lines, &opt));
     }
 
     #[test]
@@ -235,9 +244,8 @@ mod tests {
         );
         let lines: Vec<LineEntry> = vec![line1, line2, line3, line4, blank_line_entry(5, 5)];
         let expected: Vec<Warning> = vec![warning];
-        let skip_checks: Vec<LintKind> = Vec::new();
-
-        assert_eq!(expected, run(&lines, &skip_checks));
+        let opt = CliOptions::default();
+        assert_eq!(expected, run(&lines, &opt));
     }
 
     #[test]
@@ -250,9 +258,8 @@ mod tests {
         );
         let lines: Vec<LineEntry> = vec![line];
         let expected: Vec<Warning> = vec![warning];
-        let skip_checks: Vec<LintKind> = Vec::new();
-
-        assert_eq!(expected, run(&lines, &skip_checks));
+        let opt = CliOptions::default();
+        assert_eq!(expected, run(&lines, &opt));
     }
 
     #[test]
@@ -268,8 +275,8 @@ mod tests {
         ];
 
         let expected: Vec<Warning> = Vec::new();
-        let skip_checks: Vec<LintKind> = Vec::new();
 
-        assert_eq!(expected, run(&line_entries, &skip_checks));
+        let opt = CliOptions::default();
+        assert_eq!(expected, run(&line_entries, &opt));
     }
 }
--- a/dotenv-linter/src/checks.rs
+++ b/dotenv-linter/src/checks.rs
@@ -11,7 +11,7 @@ mod key_without_value;
 mod leading_character;
 mod lowercase_key;
 mod quote_character;
-mod schema_validator;
+mod schema_violation;
 mod space_character;
 mod substitution_key;
 mod trailing_whitespace;
@@ -46,7 +46,7 @@ fn checklist<'a>(opts: &'a CliOptions) -> Vec<Box<dyn Check + 'a>> {
         Box::<trailing_whitespace::TrailingWhitespaceChecker>::default(),
         Box::<unordered_key::UnorderedKeyChecker>::default(),
         Box::<value_without_quotes::ValueWithoutQuotesChecker>::default(),
-        Box::new(schema_validator::SchemaChecker::new(opts)),
+        Box::new(schema_violation::SchemaViolationChecker::new(opts)),
     ]
 }
 
--- a/dotenv-linter/src/checks.rs
+++ b/dotenv-linter/src/checks.rs
@@ -1,6 +1,6 @@
-use crate::cli::options::CliOptions;
 use crate::common::comment::Comment;
 use crate::common::{LintKind, Warning};
+use crate::schema::DotEnvSchema;
 use dotenv_lookup::LineEntry;
 
 mod duplicated_key;
@@ -31,7 +31,7 @@ pub trait Check {
 }
 
 // Checklist for checks which needs to know of only a single line
-fn checklist<'a>(opts: &'a CliOptions) -> Vec<Box<dyn Check + 'a>> {
+fn checklist<'a>(schema: Option<&'a DotEnvSchema>) -> Vec<Box<dyn Check + 'a>> {
     vec![
         Box::<duplicated_key::DuplicatedKeyChecker>::default(),
         Box::<ending_blank_line::EndingBlankLineChecker>::default(),
@@ -46,13 +46,16 @@ fn checklist<'a>(opts: &'a CliOptions) -> Vec<Box<dyn Check + 'a>> {
         Box::<trailing_whitespace::TrailingWhitespaceChecker>::default(),
         Box::<unordered_key::UnorderedKeyChecker>::default(),
         Box::<value_without_quotes::ValueWithoutQuotesChecker>::default(),
-        Box::new(schema_violation::SchemaViolationChecker::new(opts)),
+        Box::new(schema_violation::SchemaViolationChecker::new(schema)),
     ]
 }
 
-pub fn run(lines: &[LineEntry], opts: &CliOptions) -> Vec<Warning> {
-    let mut checks = checklist(opts);
-    let skip_checks = &opts.skip;
+pub fn run(
+    lines: &[LineEntry],
+    skip_checks: &[LintKind],
+    schema: Option<&DotEnvSchema>,
+) -> Vec<Warning> {
+    let mut checks = checklist(schema);
 
     // Skip checks with the --skip argument (globally)
     checks.retain(|c| !skip_checks.contains(&c.name()));
@@ -105,16 +108,16 @@ mod tests {
     fn run_with_empty_vec_test() {
         let empty: Vec<LineEntry> = Vec::new();
         let expected: Vec<Warning> = Vec::new();
-        let opt = CliOptions::default();
-        assert_eq!(expected, run(&empty, &opt));
+        let skip_checks: Vec<LintKind> = Vec::new();
+        assert_eq!(expected, run(&empty, &skip_checks, None));
     }
 
     #[test]
     fn run_with_empty_line_test() {
         let lines: Vec<LineEntry> = vec![blank_line_entry(1, 1)];
         let expected: Vec<Warning> = Vec::new();
-        let opt = CliOptions::default();
-        assert_eq!(expected, run(&lines, &opt));
+        let skip_checks: Vec<LintKind> = Vec::new();
+        assert_eq!(expected, run(&lines, &skip_checks, None));
     }
 
     #[test]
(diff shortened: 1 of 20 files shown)


pm100:
i dont understand why the code coverage is saying I am missing a test for invlaid key, the test is there, gets run and passes.

pm100:
i have fixed the missing coverages, but the coverage ci test needs to be rerun

dotenv-linter/src/checks/schema_validator.rs
```diff
@@ -0,0 +1,118 @@
...
+            last_line_number: 0,
+        }
+    }
+}
+
+impl Check for SchemaChecker<'_> {
+    fn run(&mut self, line: &LineEntry) -> Option<Warning> {
+        self.schema?;
+        self.last_line_number = line.number;
+        let schema = self.schema.unwrap();
```
DDtKey, line 31:
```suggestion
        let schema = self.schema?;
        self.last_line_number = line.number;
       
```

dotenv-linter/src/checks/schema_validator.rs
```diff
@@ -0,0 +1,118 @@
...
+        let key = line.get_key()?;
+        self.seen_keys.insert(key.to_string());
+
+        if let Some(entry) = schema.entries.iter().find(|i| i.key == key) {
```
DDtKey, line 35: Probably would be nice to represent schema elements as a hashmap?

Just to optimize the access, instead of filtering the iterator. 
Just because we do this for each line of file actually 🤔

dotenv-linter/src/schema.rs
DDtKey, line 1: Looks like a good candidate for separate sub-crate. Probably even with validation rules in it.
Another benefit - is ability to control schema versioning and align with semver of the crate

It can be considered at later period of time though

dotenv-linter/src/lib.rs
```diff
@@ -61,9 +59,18 @@ pub fn fix(opts: &FixOptions, current_dir: &PathBuf) -> Result<()> {
     let mut warnings_count = 0;
     for (index, (fe, mut lines)) in files.into_iter().enumerate() {
         output.print_processing_info(&fe);
+        // let check_options = CliOptions {
```
DDtKey, line 62: I guess just remaining after some resting

dotenv-linter/src/schema.rs
```diff
@@ -0,0 +1,310 @@
+use anyhow::Result;
+use serde::Deserialize;
+use std::{fs::File, io::BufReader, path::Path};
+#[derive(Deserialize, Default)]
+#[serde(default)]
+pub struct DotEnvSchema {
+    pub version: String,
+    pub allow_other_keys: bool,
+    pub entries: Vec<SchemaEntry>,
```
DDtKey, line 9: Yeah, looks like can be represented as a `HasMap` (name -> type & its rules)? In general it also would prevent any possible duplicate definitions

DDtKey (commented):
Thanks for the contribution! 
I really like this to be part of `dotenv-linter`! 🚀

I've left some comments for first pass

@dotenv-linter-core please take a look

DDtKey:
Another idea in my mind in continue to this one, is to allow deriving of the schema from `.env` files.

To provide a way for users to generate a schema from existing files - they can be large. And then it can be edited manually. 

**This is definitely not for this PR**. 
But highlights that it makes much more sense to extract schema logic to sub crate. It can be used even outside of `dotenv-linter`

dotenv-linter/src/checks/schema_validator.rs
pm100: I wondered about this, but could not work out how to tell serde to deseriliaze into a hashmap. I will have another pass at it.

dotenv-linter/src/checks/schema_validator.rs
pm100: lol - i originally has `if let Some()......` clippy said , replace with this i just pasted it in without thinking. fixed

pm100:
I have one failing test because the text returned by failure to open a file is different on windows (My main dev system) and linux. I will try to fix,

You code coverage tool is a hard taskmaster - but very good at making sure everything is tested.

dotenv-linter/src/checks/schema_validator.rs
DDtKey: I think it would be correct to deserialize to current `Vec` based structure and then just map all elements to hashmap, probably using enums with sub attributes (e.g validation rules for string, number and etc)

Another alternative: instead of `array`, try to use `Object` in `JSON`, where `key` is env variable name 🤔

dotenv-linter/src/schema.rs
```diff
@@ -0,0 +1,335 @@
...
+}
+#[derive(Deserialize, Default)]
+#[serde(default)]
+pub struct SchemaEntry {
+    pub key: String,
+    pub required: bool,
+    #[serde(rename = "type")]
+    pub value_type: SchemaValueType,
+    pub regex: Option<String>,
+}
```
DDtKey, line 26: Since we do conversion, what do you think of using `enums`? It also would introduce some validation on schema reading

I mean something like:
```rust
enum EnvVariableKind {
  String { regex },
  Email,
  Nunber { min, max },
  ..
}
```

dotenv-linter/src/checks/schema_validator.rs
DDtKey: Oh, I see, you already updated that!
Thanks

Left a suggestion for new code!

dotenv-linter/src/schema.rs
```diff
@@ -0,0 +1,335 @@
...
+            ]
+        }"#;
+        // write the above json to a temp file
+        let temp_dir = tempdir().unwrap();
+        let file_path = temp_dir.path().join("schema.json");
+        let schema = {
+            let mut file = File::create(&file_path).unwrap();
+            file.write_all(json.as_bytes()).unwrap();
+            // load the schema from the file
+            DotEnvSchema::load(&file_path)
```
DDtKey, line 308: I might suggest to use `tempfile` directly https://crates.io/crates/tempfile

Instead of tempdir here

dotenv-linter/src/cli.rs
```diff
@@ -130,6 +126,11 @@ fn common_args() -> Vec<Arg> {
             .long("recursive")
             .help("Recursively searches and checks .env files")
             .action(ArgAction::SetTrue),
+        Arg::new("schema")
+            .short('S')
+            .long("schema")
+            .help("Use schema file to check .env files")
+            .value_parser(value_parser!(PathBuf)),
```
DDtKey, line 133: Btw, it can be useful to have an ability to pass schema as an in-place argument (json string) to the CLI.

E.g it can be used when you're piping commands

dotenv-linter/src/schema.rs
pm100: I wondered about this too, I dont think so though. The only useful one at the moment is `String(Option<Regex>)` . If the idea is to encapsulate the entire definition of the 'rule' in the enum they get too complex. Imagine rules that say
- either KEY1 or KEY2 must be present but not both
- if KEY1 present then KEY2 must also be present
- these are the allowed values for KEY1 (could be a regex)
- ...

The code then still needs to look at other things in the schema definition , not just the type enum so my vote is no.

pm100:
I also messed up the fix for that missing file test, hopefully I have it correct now

dotenv-linter/src/schema.rs
DDtKey: If we need type-independent checks, it could be more extended wrapper, like:

```rs

entries: HashMap<String, EnvValidationRules>

struct EnvValidationRules {
   ty: EnvVariableType,
   excludes: Vec<String>, // keys which can't present in case of this one
   ... // common rules which doesn't depend on type
}

The main point here that there are inappropriate configs for particular types (i.e depends only on type). 
For now only `Regex` for string, but we already have understanding of validations for numbers and can have more.

It's kinda a good way to ensure schema is passed in correct way, and let's say there is no such confusing definitions:

```
        {
            "key": "SECRET",
            "required": true,
            "type": "Number",
            "regex": "^[a-z]*$"
        },
```

File: dotenv-linter/src/schema.rs [hunk 8]
```diff
@@ -0,0 +1,335 @@
...
+    #[serde(rename = "type")]
+    pub value_type: SchemaValueType,
+    pub regex: Option<String>,
+}
+#[derive(Deserialize, Default)]
+
+pub enum SchemaValueType {
+    #[default]
+    String,
+    Number,
```
DDtKey, line 32: Btw, I think it makes sense to be more precise for number representation here. 
E.g be able to validate integer against floating point number

dotenv-linter/src/checks/schema_validator.rs
```diff
@@ -0,0 +1,117 @@
...
+                        if value == "true" || value == "false" {
+                            return None;
+                        } else {
+                            return Some(Warning::new(
+                                line.number,
+                                self.name(),
+                                format!("The {} key is not a valid boolean", key),
+                            ));
+                        }
+                    }
```
DDtKey, line 58: I guess for boolean we also would need to think about possible validations.

Because even `clap` provide couple of parsers: 
- [BoolishValueParser](https://docs.rs/clap/latest/clap/builder/struct.BoolishValueParser.html)
- [FalseyValueParser](https://docs.rs/clap/latest/clap/builder/struct.FalseyValueParser.html)

And what to consider `true` or `false` is responsibility of application which works with that. For now we will have hard-check 🤔 

So probably we should consider passing lists of supported values.

For first version I think it's ok to stick to `true` & `false`.  I'd say lower & upper case.

DDtKey:
I apologize for such number of comments and suggestions. 
This is just a really great feature, but implementation is kinda tricky part here, it should be extensible.

Many of items can be done in iterative way, in separate PRs. Thus, we just need to collect features which we gonna schedule for the future enhancement. It also would allow other folks to contribute 🙂 

Thanks for all efforts addressing everything! 🙏

dotenv-linter/src/checks/schema_validator.rs
pm100: booleans can be done via regex too

pm100: I have updated the boolean test

pm100:
@DDtKey are you waiting for me to do something. I think I have done all that you asked, I have all checks passing and 100% coverage

DDtKey:
Actually I'm not the only maintainer, so need @dotenv-linter/core to review as well (only @mgrachev can merge / make final decision here).

But from my side, I still think it would be nice [this part](https://github.com/dotenv-linter/dotenv-linter/pull/761#discussion_r1518654043) to be implemented 🤔 
I believe it can be done on deserialization level (with [internally tagged enum](https://serde.rs/enum-representations.html#internally-tagged) where tag is type), this way we will have some basic validations of the schema structure even before applying any logic.

pm100:
That works if we are deseriazling enums directly - we are not doing that. So it has to be done manually (or at least I cant work out how to do it) - which is very simple. I can change the enum to 
```
pub enum SchemaValueType {
    #[default]
    String(Option<String>), 
    Integer,
    Float,
    Boolean,
    Url,
    Email,
}
```

if you like

I can reject a regex on anything other than string too.

Do you want that?

pm100:
I dont think serde can do it when we are not deserializing the enums directly. I can make serde do it directly from a different json syntax that I dont like so much

```
[ 
  "key":"NAME",
   "type":{"String" :null}
]
```

or 

```
[ 
  "key":"NAME",
   "type":{"String" :"regex"}
]

```
which is very odd.

I can change the enum to 

```
pub enum SchemaValueType {
    // #[default]
    String(Option<String>),
    Integer,
    Float,
    #[default]
    Boolean,
    Url,
    Email,
}
```

if you like ( I dont think it adds anything)

I will have to remap it manually - which is fine, just like the vec to hashmap translation

I will then reject a regex on anything other than a string. Or I can simply do that without changing the enum

DDtKey:
That's possible, did you consider to use `flatten`?
Here is you can find working example on rust's playground: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a272538dd8572b664bc4494804cdd3a5

mgrachev:
@pm100 Thank you for your contribution 👍

Sorry, but I need more time to review your PR, so I'll look at it around the weekend.

dotenv-linter/src/schema.rs
```diff
@@ -0,0 +1,384 @@
+use anyhow::Result;
```
DDtKey, line 1: Looks like we don't need anyhow. That's the only usecase and we can keep it simpler for now. 

I see only two possible kind of errors for now:
- io::Error (schema file handling)
- serde

Since it's limited number of errors and they have different meaning, I think it would be better to avoid extra dependency here. 
I'd suggest to use either:
- [thiserror](https://crates.io/crates/thiserror)  - it's proc (derive) macro, so dependency used only for compilation
- manually define a enum with display

I'd go with thiserror, and would create an error like that:

```rs
enum SchemaReadError {
  #[error("failed to read file: {0}")]
  Read(#[from] std::io::Error),
  #[error("schema format is invalid: {0}")]
  Invalid(#[from] serde_json::Error)
}
```

Thus, error messages will be more controllable and we avoid runtime-dependency (moreover, anyhow uses boxing)

DDtKey (changes requested):

dotenv-linter/src/schema.rs
pm100: i will take out anyhow. Its my automatic include in all projects (I think it should be in std quite honestly)

dotenv-linter/src/cli/options.rs
```diff
@@ -1,48 +1,62 @@
...
         let skip = get_many_from_args::<LintKind>(args, "skip")
             .into_iter()
             .map(|l| l.to_owned())
             .collect();
-
+        let schema = if let Some(schema_path) = args.get_one::<PathBuf>("schema") {
+            match schema::DotEnvSchema::load(schema_path) {
+                Ok(schema) => Some(schema),
+                Err(e) => {
+                    println!("Error loading schema: {}", e);
```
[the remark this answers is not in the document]
pm100: it was stderr originally but the cli tests only look at stdout

dotenv-linter/src/cli/options.rs
```diff
@@ -1,48 +1,62 @@
...
+use crate::{
+    common::LintKind,
+    schema::{self, DotEnvSchema},
+};
 use clap::ArgMatches;
 use std::path::PathBuf;
-
-pub struct CheckOptions<'a> {
+#[derive(Default)]
+pub struct CliOptions<'a> {
```
mgrachev, line 8: Since there is no need to pass the whole `CliOptions` to  the `checks::run()` function, I suggest returning everything as it was 🙂 Except for adding a new `schema` field to the `CheckOptions` structure.

dotenv-linter/tests/checks/schema.rs
```diff
@@ -0,0 +1,90 @@
...
+                "type": "Email"
+            },
+            {
+                "key": "FLAG",
+                "type": "Boolean"
+            }
+        ]
+    }"#;
+    // write the above json to a temp file
+    let temp_dir = tempdir().unwrap();
```
mgrachev, line 41: Please use `expect(msg)` instead of `unwrap()` in all places.

dotenv-linter/src/checks/schema_validator.rs
```diff
@@ -0,0 +1,138 @@
...
+        self.last_line_number = line.number;
+        let key = line.get_key()?;
+        self.seen_keys.insert(key.to_string());
+
+        if let Some(entry) = schema.entries.get(key) {
+            if let Some(value) = line.get_value() {
+                match entry.value_type {
+                    SchemaValueType::String => {
+                        if let Some(regex) = &entry.regex {
+                            if !regex::Regex::new(regex).unwrap().is_match(value) {
```
mgrachev, line 39: It doesn't seem right to parse regexp during the check and crash if it fails 🤔

dotenv-linter/src/schema.rs
```diff
@@ -0,0 +1,384 @@
...
+}
+#[derive(Deserialize, Default)]
+#[serde(default)]
+
+pub struct SchemaEntry {
+    pub key: String,
+    pub required: bool,
+    #[serde(rename = "type")]
+    pub value_type: SchemaValueType,
+    pub regex: Option<String>,
```
mgrachev, line 25:
```suggestion
    pub regex: Option<regex::Regex>,
```
I think it's best to store the already parsed regex in this field.

dotenv-linter/src/cli/options.rs
pm100: The schema is needed by fix too - thats what made me do the unifications, fix calls check first.

pm100:
All done except CliOptions - see my comment

dotenv-linter/src/cli/options.rs
mgrachev: The scheme is not needed for the `fix` command, because in any case, we cannot fix the warnings about the violation of the scheme. So, just pass `None` as schema to the `fix` function.

dotenv-linter/src/checks.rs
```diff
@@ -81,6 +88,11 @@ pub fn run(lines: &[LineEntry], skip_checks: &[LintKind]) -> Vec<Warning> {
         }
     }
 
+    for ch in &mut checks {
+        let end_warns = ch.end();
+        warnings.extend(end_warns);
+    }
```
mgrachev, line 94:
```suggestion
    if disabled_checks.contains(&LintKind::SchemaViolation) {
        return warnings;
    }

    if let Some(schema) = schema {
        let keys = lines.iter().filter_map(|l| l.get_key()).collect::<Vec<&str>>();
        let required_warnings = schema_violation::check_required_keys(schema, &keys);
        warnings.extend(required_warnings);
    }
```
It seems to me that it's more obvious what's going on.

dotenv-linter/src/checks/schema_violation.rs
```diff
@@ -0,0 +1,139 @@
...
+                    warnings.push(Warning::new(
+                        self.last_line_number,
+                        self.name(),
+                        format!("The {} key is required", entry.key),
+                    ));
+                }
+            }
+        }
+        warnings
+    }
```
mgrachev, line 138:
```suggestion
pub(crate) fn check_required_keys(schema: &crate::schema::DotEnvSchema, keys: &[&str]) -> Vec<Warning> {
    let mut warnings = Vec::new();
    for (key, entry) in &schema.entries {
        if entry.required && !keys.contains(&key.as_str()) {
            warnings.push(Warning::new(
                keys.len() + 1,
                LintKind::SchemaViolation,
                format!("The {} key is required", entry.key),
            ));
        }
    }
    warnings
}
```
The end function is not too obvious, so I propose this solution.

DDtKey: Btw, I think there is a better alternative.

- extract all required keys from schema in `SchemaViolationChecker::new` and keep as `HashSet`
- every call of `run` we remove the key from the hashset
- If at the end there are keys left in the `hashset`, we return warnings for them.

This should be more efficient than going through the keys we collected again. Also, we will not be forced to store all the keys from the file in memory.

This will still require the checker to have a callback, much like now, but instead of `end` I would suggest something like `remaining_warnings` or so. I believe this is even better, because instead of hardcode we will get expandable functionality for checkers.

pm100: the 'end' call is intended as a generic callback for all checkers that may want to do a final pass of some kind

Removing the keys once seen will prevent the validation of duplicated entries, ie regex check... The second one may be the 'intended' one

dotenv-linter/src/checks.rs
pm100: the 'end' call was intended as a generic call for any checkers that may want to do a final pass of some kind

This now elevates the schema checker to a special case

dotenv-linter/src/schema.rs
```diff
@@ -0,0 +1,484 @@
...
+    Boolean,
+    Url,
+    Email,
+}
+
+impl DotEnvSchema {
+    pub fn load(path: &Path) -> Result<Self, std::io::Error> {
+        let file = File::open(path)?;
+        let reader = BufReader::new(file);
+        let read_schema: ReadDotEnvSchema = serde_json::from_reader(reader)?;
```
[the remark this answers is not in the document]
pm100: it does not work, it did originally but once the entries was changed to HashMap and the regex was compiled into the schema at load time it is no longer possible

dotenv-linter/src/schema.rs
```diff
@@ -0,0 +1,484 @@
...
+}
+
+#[derive(Deserialize, Default)]
+#[serde(default)]
+pub struct ReadSchemaEntry {
+    pub key: String,
+    pub required: bool,
+    #[serde(rename = "type")]
+    pub value_type: SchemaValueType,
+    pub regex: Option<String>,
```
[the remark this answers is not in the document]
pm100: i did not know this one

dotenv-linter/src/schema.rs
```diff
@@ -0,0 +1,484 @@
...
+    pub entries: Vec<ReadSchemaEntry>,
+}
+
+#[derive(Deserialize, Default)]
+#[serde(default)]
+pub struct ReadSchemaEntry {
+    pub key: String,
+    pub required: bool,
+    #[serde(rename = "type")]
+    pub value_type: SchemaValueType,
```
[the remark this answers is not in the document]
pm100: I prefer not to do that

dotenv-linter/src/cli/options.rs
pm100: @mgrachev  schema can be used by both fix and compare.

- fix can remove (or better, comment out) lines that should not be there
- compare can use the schema , 123.0, 123 and 0123 are all the same if type is double

dotenv-linter/src/checks/schema_violation.rs
DDtKey: > Removing the keys once seen will prevent the validation of duplicated entries, ie regex check... The second one may be the 'intended' one

Didn't get it. What do you mean?

I expect to have something like `required_keys: HashSet<String>` and we remove the keys from it once seen. It shouldn't be used for anything but required keys

Required key present - we remove it from the set.  The logic is almost the same as current one with second pass through all keys.

dotenv-linter/src/schema.rs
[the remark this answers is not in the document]
mgrachev: Why?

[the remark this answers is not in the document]
DDtKey: It's quite subjective topic, I also usually prefer `rename` over `r#` prefix, e.g:
```rs
#[serde(rename = "type")]
ty: Type,
```

`r#type` is not the most convenient and understandable when it comes to code in which you need to access the field

dotenv-linter/src/schema.rs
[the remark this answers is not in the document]
DDtKey: Yeah, I'm good with that. Actually I mentioned this way [earlier](https://github.com/dotenv-linter/dotenv-linter/pull/761#discussion_r1518575976):
> Also alternative to consider: instead of `array`, try to use `Object` in `JSON`, where `key` is env variable name 🤔 This can be deserialized into map

Any valid Unicode string is valid json key - so we should have no issues with that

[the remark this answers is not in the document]
DDtKey: Another "unresolved" question regarding schema deserialization part is https://github.com/dotenv-linter/dotenv-linter/pull/761#issuecomment-1991243721 (also there is an example in rust-playground)

I tend to keep type-specific rules attached to the type itself.

pm100:
reverted all Cli changes

using serde_regex

dotenv-linter/src/schema.rs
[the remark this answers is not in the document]
mgrachev: > Another "unresolved" question regarding schema deserialization part is [#761 (comment)](https://github.com/dotenv-linter/dotenv-linter/pull/761#issuecomment-1991243721) (also there is an example in rust-playground)
> 
> I tend to keep type-specific rules attached to the type itself (like `regex` belongs to `string`). Just to keep our code more extensible & clear for the future

I like your idea 👍

pm100:
Why not accept the PR, which is fully featured and passes 100% test coverage - as far as I all the proposed changes are stylistic choices - then change it however you like. There are tons of extra features that can be added once the base support is in:
- use the schema for fix and compare
- add numeric ranges
- add min and max lengths
- add allowed values
- add support for dir and file type , both syntactically valid and optionally check for existence
- improve booleans (localizable???)

Or close it

DDtKey:
I think there is no point in merging the PR with an unfinished schema. This is the most important part of PR.
And if we accept it unfinished, then we need someone to work on it to get the job done. 

We are not asking for additional features, just changes to an important part. I don't think merging the intermediate state of a feature is a good idea.

Even if we did, we wouldn't be able to release it until we were ready to introduce the schema-format without making incompatible changes immediately afterwards.

But you have already done a huge amount of work (thank you!) and are familiar with how this can be changed

pm100:
I do not know what changes to the schema are being asked, I only see implementation changes. Can somebody please post the proposed schema syntax and I will implement it

DDtKey:
Sure, the suggestion was in the thread above https://github.com/dotenv-linter/dotenv-linter/pull/761#discussion_r1531048360

pm100:
the problem with that schema is that it is not describable in a json schema itself, so json aware editors wont like it.

The only reason for the proposed change is to slightly simplify the implementation on reading the schema. Which isnt a compelling reason IMHO

DDtKey:
> the problem with that schema is that it is not describable in a json schema itself, so json aware editors wont like it.

Didn't really get your point. It's valid JSON and I any JSON editor works with this.

> The only reason for the proposed change is to slightly simplify the implementation on reading the schema. Which isnt a compelling reason IMHO

Well, I see at least couple of reasons:
- any JSON linter/validator will highlight duplicative keys. We won't have to deal with this explicitly
- It has a good representation in yaml (we may want to support yaml or toml in the future), for example above:

```yaml
---
---
version: '1.0'
allow_other_keys: false
entries:
  SECRET:
    required: true
    type: String
    regex: "^[a-z]*$"
  PORT:
    type: Integer
  URL:
    type: Url

```

pm100:
ok, I can change to that format. But its much harder to detect duplicated keys - which I have to do and forgot in the original code.

The problem is that serde happily accepts duplicates. Ie this

```
{
    "version": "1.0.0",
    "allow_other_keys": false,
    "entries": {
        "SECRET": {
            "required": true,
            "type": "String",
            "regex": "^[a-z]*$"
        },
        "PORT": {
            "type": "String"
        },
        "PORT": {
            "type": "Integer"
        }
    }
}
```
It means that I have to deserialize some other way

see https://stackoverflow.com/questions/64405600/how-can-i-deserialize-json-that-has-duplicate-keys-without-losing-any-of-the-val

And since the only real purpose of this change was to make the deserialze easier I dont really like it.

BTW - it is possible tp describe this format using json schema - it just wasnt obvious

DDtKey:
> The problem is that serde happily accepts duplicates.

Most common solution is to use `serde_with` for this:
https://docs.rs/serde_with/latest/serde_with/rust/maps_duplicate_key_is_error/index.html

That's why I mentioned that we can get duplicate-check for free.

> And since the only real purpose of this change was to make the deserialze easier I dont really like it.

Please don't claim that this is the only reason. I suggested this earlier and above you see at least two more reasons

pm100:
>That's why I mentioned that we can get duplicate-check for free.

we get duplicate check for free in some editors, not at run time. So with this schema json dotenv-linter will accept schemas with duplicated keys, which is not what you want (I assume)

DDtKey:
> we get duplicate check for free in some editors, not at run time. So with this schema json dotenv-linter will accept schemas with duplicated keys, which is not what you want (I assume

Sorry, but did you read what I suggested? (don't consider it rude) 
There is a solution to disallow duplicative keys on deserialization step with `serde_with`

pm100:
done

mgrachev (approved):

DDtKey (approved):
LGTM
Couple of iterations before releasing should be performed

mgrachev:
@pm100 Thank you for your help 👍

Issue #776, opened 2024-04-23
Expand syntax support for variable default and alternate values


nogweii:
I have a bunch of [systemd environment.d](https://www.freedesktop.org/software/systemd/man/latest/environment.d.html) files, and it supports a couple of additional bash syntaxes for setting a default value (with a colon and dash) and an alternate value (with a colon and plus).

Example:

```sh
FOO=${NODE_ENV:-development if not set}
BAR=${CI:+only when running in CI}
```

And right now, using this syntax results in `SubstitutionKey` errors.

mgrachev:
@nogweii Hi! Thank you for your feedback!

I think we should support this syntax too.

SkyWarrior123:
Hi @mgrachev , can I work on the above to add the syntax  ?

R11baka:
@mgrachev Created mr for fix this issue. https://github.com/dotenv-linter/dotenv-linter/pull/786

hriztam:
Hello, if the issue is still open, I would like to work on it

luccinmasirika:
Hi 👋

I'd like to pick this up. The existing PR #786 looks inactive since mid-2024, and the changes @DDtKey requested there were never applied.

I've implemented it following exactly the validation approach discussed in that PR:

- Closed list of allowed operators: `:-`, `:+`, `:=`, `:?` (no open-ended regex)
- Each operator may appear **at most once** per substitution
- Multiple separate substitutions are accepted, e.g. `${NODE_ENV:-development}${CI:+only}`
- Invalid inputs like `${CI:+one:-two:?three}` and `${SOME_VALUE--$$++???_END}` are still rejected

While testing, I also found a related bug: `get_substitution_keys` was including the operator in the extracted key (e.g. `BAR:-default` instead of `BAR`), which produced false `UnorderedKey` positives once this syntax is allowed. I fixed that too, with tests.

Would you be open to a PR? I'll credit @R11baka for the original work in #786.

PR #786, opened 2024-05-31
fix(749): allow alternate syntax for bash variables

R11baka:
#### ✔ Checklist:

- [x ] Tests for the changes have been added (for bug fixes / features);

### DIFF
--- a/dotenv-linter/src/checks/substitution_key.rs
+++ b/dotenv-linter/src/checks/substitution_key.rs
@@ -44,7 +44,11 @@ impl Check for SubstitutionKeyChecker<'_> {
             let is_incorrect_substitution = has_start_brace ^ has_end_brace
                 || end_brace_index
                     .map(|i| &initial_key[1..i])
-                    .filter(|key| key.contains(|c: char| !c.is_ascii_alphanumeric() && c != '_'))
+                    .filter(|key| {
+                        key.contains(|c: char| {
+                            !c.is_ascii_alphanumeric() && !":-+=?".contains(c) && c != '_'
+                        })
+                    })
                     .is_some();
 
             if is_incorrect_substitution && !is_escaped(prefix) {
@@ -136,6 +140,19 @@ mod tests {
                 ("FOO=\\$BAR}", None),
                 ("FOO=\"\\${BAR\"", None),
                 ("FOO=\"\\$BAR}", None),
+            ],
+        );
+    }
+
+    #[test]
+    fn check_variable_alternate() {
+        check_test(
+            &mut SubstitutionKeyChecker::default(),
+            [
+                ("FOO=${NODE_ENV:development}", None),
+                ("BAR=${CI:+only}", None),
+                ("BUS=${var:=DEFAULT}", None),
+                ("FOOBAR=${var:?IS_UNSET}", None),
             ],
         );
     }


dotenv-linter/src/checks/substitution_key.rs
```diff
@@ -139,4 +143,17 @@ mod tests {
...
+
+    #[test]
+    fn check_variable_alternate() {
+        check_test(
+            &mut SubstitutionKeyChecker::default(),
+            [
+                ("FOO=${NODE_ENV:development}", None),
+                ("BAR=${CI:+only}", None),
+                ("BUS=${var:=DEFAULT}", None),
+                ("FOOBAR=${var:?IS_UNSET}", None),
```
mgrachev, line 155:
```suggestion
                ("FOOBAR=${var:?IS_UNSET}", None),
                ("BUZZ=${BAR:-DEFAULT}", None),
```
Let's add another example.

mgrachev (commented):
👋 Thank you for your contribution 👍

DDtKey:
Just one a nitpick: I would consider this as a `feature` rather than `fix`. 
Because we introduce support for alternative syntax.

dotenv-linter/src/checks/substitution_key.rs
```diff
@@ -44,7 +44,11 @@ impl Check for SubstitutionKeyChecker<'_> {
             let is_incorrect_substitution = has_start_brace ^ has_end_brace
                 || end_brace_index
                     .map(|i| &initial_key[1..i])
-                    .filter(|key| key.contains(|c: char| !c.is_ascii_alphanumeric() && c != '_'))
+                    .filter(|key| {
+                        key.contains(|c: char| {
+                            !c.is_ascii_alphanumeric() && !":-+=?".contains(c) && c != '_'
```
DDtKey, line 49: `!":-+=?".contains(c) && c != '_'` 

Looks like it can be merged to `!":-+=?_".contains(c)`

DDtKey: another suggestion:
We could define a static `HashSet` to slightly improve the perf.  However it requires either LazyCell (stabilized recently) or lazy_static

E.g:

```rust
// or `LazyLock`
const ALLOWED_CHARS: LazyCell<HashSet<char>> = LazyCell::new(|| HashSet::from([':', '-', '+', '=', '?', '_']));
// ...
&& !ALLOWED_CHARS.contains(c)
```

dotenv-linter/src/checks/substitution_key.rs
```diff
@@ -139,4 +143,18 @@ mod tests {
...
+
+    #[test]
+    fn check_variable_alternate() {
+        check_test(
+            &mut SubstitutionKeyChecker::default(),
+            [
+                ("FOO=${NODE_ENV:development}", None),
+                ("BAR=${CI:+only}", None),
+                ("BUS=${var:=DEFAULT}", None),
+                ("FOOBAR=${var:?IS_UNSET}", None),
```
DDtKey, line 155: Btw, @mgrachev do you think it's ok just to allow these chars?

#776 syntax still has some rules, while we may encounter something like: `FOO=SOME_VALUE--$$++???_END` - which is invalid

dotenv-linter/src/checks/substitution_key.rs
DDtKey: But hold on, please
First of all, we need to figure out our expectations here: https://github.com/dotenv-linter/dotenv-linter/pull/786/files#r1625018876

DDtKey: But hold on, please
First of all, we need to figure out our expectations here: https://github.com/dotenv-linter/dotenv-linter/pull/786/files#r1625018876

DDtKey (changes requested):

dotenv-linter/src/checks/substitution_key.rs
mgrachev: You're right, we definitely shouldn't skip such values: `FOO=${SOME_VALUE--$$++???_END}`.

Perhaps we can check such values using a regular expression of the form `\:[\+\=\?\-]?`?

DDtKey: Regex may do the trick, but we also need to ensure it's only 1 time occurrence, and not something like:
`${SOME_VALUE:+one:-two:?three}`

It still possible ofc, but probably we can just go with simpler solution: 
- pre-define the list of allowed sequences, e.g:  `[":+", ":-", ":=", ":?"]` 
- ensure there is only one occurrence within the value


Also, need to ensure it will work with multiple substitutions correctly, e.g:
- positive test: `FOO=${NODE_ENV:development}${CI:+only}`
- negative test: FOO=${NODE_ENV:development}${CI:+only:?IS_UNSET:other}`

dotenv-linter/src/checks/substitution_key.rs
DDtKey: These suggestions don't make sense, because the approach should be changed

dotenv-linter/src/checks/substitution_key.rs
mgrachev: Good idea 👍

DDtKey:
Hi @ro0NL 👋

First of all, requested changes must be applied: https://github.com/dotenv-linter/dotenv-linter/pull/786#discussion_r1625691842

Otherwise the PR will introduce another issue.

@R11baka any chance you will take a look / work on this? Just to understand the state of PR

Thanks everyone 🙏

PR #835, opened 2024-11-17
Issue 808: inline comments fix

Turtel216:
The ValueWithoutQuotesChecker did not recognize inline comments as correct. I removed the comments before checking for invalid quotes

#### ✔ Checklist:

- [ ] Commit messages have been written in [Conventional Commits](https://www.conventionalcommits.org) format;

### DIFF
--- a/dotenv-linter/src/checks/value_without_quotes.rs
+++ b/dotenv-linter/src/checks/value_without_quotes.rs
@@ -55,6 +55,8 @@ mod tests {
                 ("FOO=BAR BAZ", Some(WARNING)),
                 ("FOO=\"BAR BAZ\"", None),
                 ("FOO=\'BAR BAR\'", None),
+                ("FOO=BAR # Some Comment", None),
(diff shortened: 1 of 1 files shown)


dotenv-linter/src/checks/value_without_quotes.rs
```diff
@@ -55,6 +61,8 @@ mod tests {
                 ("FOO=BAR BAZ", Some(WARNING)),
                 ("FOO=\"BAR BAZ\"", None),
                 ("FOO=\'BAR BAR\'", None),
+                ("FOO=BAR # Some Comment", None),
+                ("FOO=BAR BAZ # Some Comment", Some(WARNING)),
```
ccoVeille, line 65: Are your changes safe with something like this ?

```
FOO="Pas#word"
FOO="Pa #word"
```

Please add tests

Turtel216: I have updated both the fix and the tests in order to cover these test cases with my recent commits

PR #949, opened 2026-03-05
fix(substitution_key): do not auto-fix ambiguous braced cases (fixes #918)

### LINKED ISSUE #918: Bug: SubstitutionKeyFixer creates malformed value with extra brace
### Description

When the `SubstitutionKeyFixer` runs on a variable that contains invalid characters within curly braces (e.g., `${VAR-NAME}`), it produces a malformed string by adding an extra closing brace `}`.

### Steps to Reproduce

1.  Create a `.env` file with the following content. This line contains an invalid character (`-`) inside the braces, which triggers the `SubstitutionKey` checker.
    ```dotenv
    URL=${FOO-BAR}://$HOST-db
    ```

2.  Run the `fix` command:
    ```bash
    dotenv-linter fix
    ```

### Current Behavior

The `dotenv-linter fix` command reports a `SubstitutionKey` warning and attempts to fix the line. However, it produces an incorrect string with a misplaced closing brace.

The file content is changed to:

```dotenv
URL=${FOO}-BAR}://${HOST}-db

Nichokas:
Fixes #918 
Leaves `${FOO-BAR}` untouched to prevent the extra `}` bug (as suggested by @DDtKey).  
Now it only fixes bare `$KEY` cases.

#### ✔ Checklist:

### DIFF
--- a/dotenv-analyzer/src/fix/substitution_key.rs
+++ b/dotenv-analyzer/src/fix/substitution_key.rs
@@ -25,6 +25,15 @@ impl Fix for SubstitutionKeyFixer {
             result.push_str(prefix);
             result.push('$');
 
+            if raw_key.starts_with('{') {
+                if let Some(closing_pos) = raw_key.find('}') {
+                    let whole_sub = &raw_key[..=closing_pos];
+                    result.push_str(whole_sub);
+                    value = &raw_key[closing_pos + 1..];
+                    continue;
+                }
+            }
+
             // Separate initial key from the rest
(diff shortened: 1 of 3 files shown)


mstruebing:
@Nichokas I've started CI and it seems like some tests are failing as well as one if statement that needs to be collapsed somehow.

Nichokas:
@mstruebing Thanks for the review

I'm new to the project, so I missed both the CLI integration test and the Clippy suggestion. Just pushed the fixes.

Issue #749, opened 2024-01-30
on fix action support comment for group


m47730:
Hi,
first of all thanks for this tool, i love it!

I struggle with this: if i try to automatically fix an unsorted env files with groups with comment as delimiter

for example:
```
ENV2=bbb
ENV1=aaa

###> group1 ###
ENV4=ddd
ENV5=eee
ENV3=ccc
###< group1 ###

ENV7=ggg
ENV6=fff
```

ends up with a sorted list but some key go outside the "comment group" (see `ENV3`):

```
ENV1=aaa
ENV2=bbb

ENV3=ccc
###> group1 ###
ENV4=ddd
ENV5=eee
###< group1 ###

ENV6=fff
ENV7=ggg
```

instead i'd like this output:
```
ENV1=aaa
ENV2=bbb

###> group1 ###
ENV3=ccc
ENV4=ddd
ENV5=eee
###< group1 ###

ENV6=fff
ENV7=ggg
```

mgrachev:
@m47730 Hi 👋

Thank for your feedback! It's a bug, we'll fix it.

DDtKey:
@mgrachev I don't think it's a bug

Comment is relates to ENV variable and not to group. AFAIR it was expected behavior.
```env
###> group1 ###
ENV4=ddd
```
So. we keep it after reordering for the initial ENV variable. Otherwise how to understand if comment relate to group or env variable?


But we could consider to keep comments in case of empty line above or below, probably with ability to control the behavior.

m47730:
Thank you for the workaround (it works as expected)

Could I suggest an option for unbind the comment to next line?
Something like `--do-not-move-comments` or similiar

DDtKey:
I think it makes sense to have an argument for this behavior

reubenborrego:
Saw on your repo description that you're open to newcomers.  I'm about as green to group development as can be.  Though I do have plenty of experience in Rust and development overall (And plenty of time to spare).

I wouldn't mind trying to add this QOL improvement.  All I need is a bit of upfront guidance and I could probably throw it together.  In the meantime I'll keep poking through the repo until I feel confident enough to try something on my own.

reubenborrego:
The comment appears to be moved in UnorderedKeyFixer's sort_part function.  Altering the the `!line.is_comment()` condition seemed to work well enough.

[...]

PR #954, opened 2026-04-29
fix: keep comment-delimited groups intact in unordered key fixer

hriztam:
Fixes `UnorderedKey` auto-fix for comment-delimited env groups. ( #749 )

Previously, markers like `###> group1 ###` could move with the next key during sorting, which allowed keys to escape the intended group. This change keeps those marker comments fixed and sorts keys only within the delimited section.

Added regression tests for both the fixer and the CLI behavior.

#### ✔ Checklist:


mstruebing:
I've started the workflow runs seems like you need to adjust your code a litte bit.

hriztam:
Ok let me fix it

hriztam:
I fixed it let me know if the issue is still there @mstruebing

PR #962, opened 2026-09-07
Added Harbor install support

kauanmezavila:
Created HarborSpecs folder and HarborMap.yaml for harbor install support, and added 'install with harbor' session method on README.md


kauanmezavila:
Forget, i already created a other way

